PR #53, opened 2021-12-14
Add Directory.Build.props and update to Avalonia 0.10.10

wieslawsoltes:


jinek:
@wieslawsoltes  Thank you for the update.
I see, some things started to work now, like "null object" example
image

wieslawsoltes:
> @wieslawsoltes Thank you for the update. I see, some things started to work now, like "null object" example image

Nice!

There is constant flow of fixes into Avalonia, latest bits should also support .NET 6 and webassembly. I am waiting for 0.10.11-rc.2 to hit final release (waiting for skiasharp stable release with fixes)

Also now with that PR merged you can do `dotnet pack -c Release` to create NuGet packages.

PR #58, opened 2022-03-11
contributing.md introduced

jinek:
Tried to keep the description as simple as possible.
Things, like code style/formatting are defined by DotSettings and editorconfig. Otherwise resharper is supposed to be used for formatting locally or on CI.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,0 +1,18 @@
+# Contributing Guide
+
+## Preparation
+To compile and debug the application it's necessary to install [dotnet 5 sdk](https://dotnet.microsoft.com/en-us/download/dotnet/5.0). It's possible to use any compliant tools for coding and debugging.
+
+## Tickets And Discussions
+Feel free to open a [ticket](https://github.com/jinek/Consolonia/issues) or a [discussion](https://github.com/jinek/Consolonia/discussions) in case you have issues, need help, or just have some ideas.
+
+In case of a bug, first try to check whether issue exist in [AvaloniaUI](https://github.com/AvaloniaUI/Avalonia) - Avalonia community also accepts tickets, we should consider fixing it in Avalonia rather than in Consolonia if applicable.
+
+## Code Style
+Projects are setup to utilize [dotnet analyzers](https://docs.microsoft.com/en-us/dotnet/framework/code-analyzers).
+All warning are considered as errors.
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -0,0 +1,18 @@
+# Contributing Guide
+
+## Preparation
+To compile and debug the application it's necessary to install [dotnet 5 sdk](https://dotnet.microsoft.com/en-us/download/dotnet/5.0). It's possible to use any compliant tools for coding and debugging.
```
vabka, line 4: Are there any plans for .NET 6?

contributing.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Tickets And Discussions
+Feel free to open a [ticket](https://github.com/jinek/Consolonia/issues) or a [discussion](https://github.com/jinek/Consolonia/discussions) in case you have issues, need help, or just have some ideas.
+
+In case of a bug, first try to check whether issue exist in [AvaloniaUI](https://github.com/AvaloniaUI/Avalonia) - Avalonia community also accepts tickets, we should consider fixing it in Avalonia rather than in Consolonia if applicable.
+
+## Code Style
+Projects are setup to utilize [dotnet analyzers](https://docs.microsoft.com/en-us/dotnet/framework/code-analyzers).
+All warning are considered as errors.
+Formatting and styling is set by editorconfig.yml which can be recognized by [JetBrains Resharper or Rider](https://www.jetbrains.com/resharper/). If you don't have one, just make a pull request and check build statuses. It does automatic formatting and checks the style.
```
vabka, line 14: editorconfig also recognizable by [dotnet format](https://github.com/dotnet/format) - it's a great tool if you do not have Rider or something went wrong with IDE.

Also I need we need to consider about adding dotnet format to git/hooks

contributing.md
jinek: Not planned yet, but should be simple to implement

contributing.md
jinek: I have tried to format with "dotnet format", it works differently from Resharper. Either config needs to be adjusted or "dotnet format" itself fixed.

vabka: Resharper has its own properties in editorconfig.

I think it can be fixed in another PR

vabka:
Lgtm

PR #69, opened 2022-04-18
Refactoring #6 

jinek:
Startup methods re-arranged


jinek:
Failed to build because of https://github.com/ad-m/github-push-action/issues/113

ZPascal:
@jinek You should switch the used version to of the GitHub push action to `ad-m/github-push-action@master`. That should solve your issue, because the fix is already implemented [there](https://github.com/ad-m/github-push-action/commit/6defe3bd9f0a0846636a7d99ab96ccc75f2c858b).

Issue #132, opened 2024-11-01
Underline goes 1 character to far


tomlm:
When you user underline

```xaml
    <Grid>
        <TextBlock TextDecorations="Underline">This is a test</TextBlock>
    </Grid>
```
it generates 1 char to much
![image](https://github.com/user-attachments/assets/35657056-df37-4858-b619-dbfbe16ef1fc)

tomlm:
Also, Line has same bug
```xaml
<StackPanel>
    <TextBlock Foreground="White">0123456789</TextBlock>
    <Canvas Grid.Row="1" Background="Black" Width="20" Height="1" HorizontalAlignment="Left" >
        <Line StartPoint="0,0" EndPoint="10,0" Stroke="Yellow" StrokeThickness="1"/>
    </Canvas>
</StackPanel>
```
draws one char too many
![image](https://github.com/user-attachments/assets/47d7bedb-f9fc-4e1c-91d7-51cc54b04f7e)

PR #133, opened 2024-11-01
Add Italic, strikethrough and fix line draw bug 

tomlm:
* Added Italic support
* Turned on Strikethrough (although it renders as underline)
* Added Canvas gallery page
* Removed dependency on Crayon library

Bug Fixes
* Fix line draws 1 too many chars #132 

Changes:
* Remove dependency on Crayon library, as it didn't implement italics and added ConsoleUtils
* Use Italic escape to handle FontStyle=Italic
* Refactor DrawLineInternal into  DrawLineInternal and DrawRectangleLineInternal

Details:
The issue is that DrawLineInternal was written to draw rectangle borders and then simply reused as the DrawLine implementation. The original DrawLineInternal draws Corner + Line + Corner. This always ended up drawing 1 more than the desired line length, when reused for a simple line.

The solution was to refactor the code into
* DrawLineInternal
* DrawRectangleLineInternal

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `GalleryCanvas` user control for enhanced graphical representation, including lines and rectangles.
	- Added text decoration demonstrations in `GalleryTextBlock`, showcasing underline, strikethrough, and italic styles.
	- New utility class `ConsoleUtils` for ANSI escape code definitions and text formatting.

- **Bug Fixes**
	- Improved text rendering logic in `InputLessDefaultNetConsole`, allowing for more flexible formatting.

- **Refactor**
	- Enhanced drawing functionality and error handling in `DrawingContextImpl`.
	- Updated strikethrough metrics in `GlyphTypeface`.

- **Chores**
	- Removed unused package reference for `Crayon`.
	- Removed metadata attributes from `GalleryColors`, `GalleryGradientBrush`, and `GalleryImage` classes.
	- Removed references to unused XAML and code files in the project configuration.

### DIFF
--- a/src/Consolonia.Core/Consolonia.Core.csproj
+++ b/src/Consolonia.Core/Consolonia.Core.csproj
@@ -6,7 +6,6 @@
         <PackageReference Include="Avalonia.FreeDesktop" Version="$(AvaloniaVersion)" />
         <PackageReference Include="Avalonia.Skia" Version="$(AvaloniaVersion)" />
         <PackageReference Include="Avalonia.Controls.DataGrid" Version="$(AvaloniaVersion)" />
-        <PackageReference Include="Crayon" Version="2.0.69" />
         <PackageReference Include="NullLib.ConsoleEx" Version="1.0.4.4" />
         <PackageReference Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.88.8" />
     </ItemGroup>
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -146,7 +146,7 @@ namespace Consolonia.Core.Drawing
                         CurrentClip.ExecuteWithClipping(new Point(px, py), () =>
                         {
                             _pixelBuffer.Set(new PixelBufferCoordinate((ushort)px, (ushort)py),
-                                (pixel, bb) => { return pixel.Blend(new Pixel(new PixelBackground(bb.Mode, bb.Color))); },
+                                (pixel, bb) => pixel.Blend(new Pixel(new PixelBackground(bb.Mode, bb.Color))),
                                 backgroundBrush);
                         });
                     }
@@ -289,7 +289,6 @@ namespace Consolonia.Core.Drawing
 
             byte pattern = (byte)(line.Vertical ? 0b1010 : 0b0101);
             DrawPixelAndMoveHead(ref head, line, lineStyle, pattern, color, line.Length); //line
-            return;
         }
 
         /// <summary>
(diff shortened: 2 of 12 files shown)


jinek:
@tomlm don't you know why "code inspection" build did not work in this PR ?

jinek:
@tomlm  Is this correct drawing? 
image

tomlm:
There should be an orange box pinned to lower right corner....

jinek:
> There should be an orange box pinned to lower right corner....

This one I've tried Fedora+Gnome Terminal

tomlm:
Hmmm, I'll try on Ubuntu proper, I was using Ubuntu on wsl

jinek:
> There should be an orange box pinned to lower right corner....
> 
> The upper left is testing horizontal and vertical one char lines
> 
> The bottom right is testing long lines that overlap
> 
> Lines look correct, but I don't know where the orange rect is

oh ok, it just did not fit:

Screenshot from 2024-11-02 00-13-44

tomlm:
I was thinking about this more and I'm not certain I'm doing this right.

If I understand the current code, the goal is that lines actually meet at corners.

Effectively, startChar and endChars are lines that end at the middle of the pixel, so when you merge them the corners align.

So if I draw a line from pixel 0->pixel 1, what should it draw? It actually seems like it should draw a DOT at the corner. That seems weird.

But right now with my delta it does the following:
* DrawLineInternal => Draws from edge to edge.
* DrawRectangleLineInternal => Draws from center to center.
That's why my lines make overlapping cross where they pass. 

Thoughts?

jinek:
I'm not sure I properly understand your concern.

> I was thinking about this more and I'm not certain I'm doing this right.
> 
> If I understand the current code, the goal is that lines actually meet at corners.
> 
> Effectively, startChar and endChars are lines that end at the middle of the pixel, so when you merge them the corners align.

Correct.

> So if I draw a line from pixel 0->pixel 1, what should it draw? It actually seems like it should draw a DOT at the corner. That seems weird.
> 
> But right now with my delta it does the following:
> 
>     * DrawLineInternal => Draws from edge to edge.
> 
>     * DrawRectangleLineInternal => Draws from center to center.
>       That's why my lines make overlapping cross where they pass.
> 
> 
> Thoughts?

Similar thoughts, but for now I can not fully understand how the line drawing is gonna be used. Will it be the line aligned to the text, like this?

```
────
The text.
```
Or will it be the line aligned to the rectangle, like this

```
╶─╴
┌─┐
│    │
└─┘

```

I would say we just need more input in the future on this question.

jinek (approved):

tomlm:
I'm thinking we should extend line rush to have a policy for character layout

- Edge or center point
- perhaps alignment top/center/bottom assuming we can honor them with appropriate characters

Sometime like that...

PR #152, opened 2024-11-20
Upgrade Avalonia to 11.2.1

tomlm:
* I bumped version to 11.2.1
* I updated signature for changed methods
* implemented stubs for called methods which seem ok to ignore
* have not ImplementedException stubs for things I don't know if we need or not.
* Fixed some things that are obselete (RadiusX and RadiusY instead of Radius for example)

Everything seems to work except
* Menu
* ComboBox
* Dialog

They are blowing up inside KeyboardNavigationHandler, I don't know what those are/do, so I am pushing this branch up there. Maybe the solution is obvious to you?

### DIFF
--- a/src/Consolonia.Core/Drawing/ConsoloniaPlatformRenderInterfaceContext.cs
+++ b/src/Consolonia.Core/Drawing/ConsoloniaPlatformRenderInterfaceContext.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Avalonia;
 using Avalonia.Platform;
 
 namespace Consolonia.Core.Drawing
@@ -20,6 +21,13 @@ namespace Consolonia.Core.Drawing
             return new RenderTarget(surfaces);
         }
 
+        public IDrawingContextLayerImpl CreateOffscreenRenderTarget(PixelSize pixelSize, double scaling)
+        {
+            throw new NotImplementedException();
+        }
+
         public bool IsLost => false;
+
+        public IReadOnlyDictionary<Type, object> PublicFeatures => new Dictionary<Type, object>();
     }
 }
--- a/src/Consolonia.Core/Drawing/RenderTarget.cs
+++ b/src/Consolonia.Core/Drawing/RenderTarget.cs
@@ -71,6 +71,8 @@ namespace Consolonia.Core.Drawing
 
         public IDrawingContextImpl CreateDrawingContext(bool useScaledDrawing)
         {
+            if (useScaledDrawing == true)
+                throw new NotImplementedException($"Consolonia doesn't support useScaledDrawing");
             return new DrawingContextImpl(_consoleWindow, _bufferBuffer);
         }
 
--- a/src/Consolonia.Core/Drawing/RenderTarget.cs
+++ b/src/Consolonia.Core/Drawing/RenderTarget.cs
@@ -69,7 +69,7 @@ namespace Consolonia.Core.Drawing
 
         bool IDrawingContextLayerImpl.CanBlit => true;
 
-        public IDrawingContextImpl CreateDrawingContext()
+        public IDrawingContextImpl CreateDrawingContext(bool useScaledDrawing)
         {
             return new DrawingContextImpl(_consoleWindow, _bufferBuffer);
         }
--- a/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
+++ b/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
@@ -152,7 +152,6 @@ namespace Consolonia.Core.Infrastructure
         public Size MaxAutoSizeHint { get; }
 
         // ReSharper disable once UnassignedGetOnlyAutoProperty todo: what is this property
-        public IScreenImpl Screen => null!;
 
         public void SetTitle(string title)
         {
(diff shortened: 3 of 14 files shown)


src/Consolonia.Core/Drawing/ConsoloniaPlatformRenderInterfaceContext.cs
```diff
@@ -20,6 +21,13 @@ public IRenderTarget CreateRenderTarget(IEnumerable<object> surfaces)
...
         }
 
+        public IDrawingContextLayerImpl CreateOffscreenRenderTarget(PixelSize pixelSize, double scaling)
+        {
+            throw new NotImplementedException();
+        }
+
         public bool IsLost => false;
+
+        public IReadOnlyDictionary<Type, object> PublicFeatures => new Dictionary<Type, object>();
```
jinek, line 31: May be lets replace with just {get;} = new Dictionary<...>();  ? Just not to create an object on each call?

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -69,7 +69,7 @@ void IDrawingContextLayerImpl.Blit(IDrawingContextImpl context)
 
         bool IDrawingContextLayerImpl.CanBlit => true;
 
-        public IDrawingContextImpl CreateDrawingContext()
+        public IDrawingContextImpl CreateDrawingContext(bool useScaledDrawing)
```
jinek, line 72: Shouldn't we throw NotImplemented when argument is true? Just to notify we do not support scaledDrawing.

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -76,7 +76,7 @@ public IPopupImpl CreatePopup()
 
         public void SetTransparencyLevelHint(IReadOnlyList<WindowTransparencyLevel> transparencyLevels)
         {
-            throw new NotImplementedException("Consider this");
+            // I think we can ignore this.
```
jinek, line 79: Could you leave info when this happens? If well known

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -134,7 +134,7 @@ public void Activate()
 
         public void SetTopmost(bool value)
         {
-            throw new NotImplementedException();
+            // todo
```
jinek, line 137: Leave please a comment when it was called

src/Consolonia.Themes.TurboVision/Templates/Controls/Helpers/ConsoloniaTextPresenter.cs
```diff
@@ -71,10 +71,14 @@ public ConsoloniaTextPresenter()
         {
             // we need to disable blinking caret, our terminal caret blinks itself once shown
             var caretTickTimer = (DispatcherTimer)TickTimerField.GetValue(this);
-            caretTickTimer!.Interval =
-                TimeSpan.FromMilliseconds(int
-                    .MaxValue); //see DispatcherTimer.Interval, since we can not disable it, setting it to the longest interval possible
-            caretTickTimer!.Tick += (_, _) => throw new NotImplementedException("How to disable timer completely?");
+            // TODO: consider that with avalonia 11.1.5 this is null;
```
jinek, line 74: Probably we should investigate how to disable blinking in Avalonia 11.1.5 ?

src/Consolonia.Core/Drawing/ConsoloniaPlatformRenderInterfaceContext.cs
tomlm: This is a shorthand for that:
  private string _name;  // the name field
  public string Name => _name;     // the Name property with get only accessor

jinek: I mean
```
public IReadOnlyDictionary<Type, object> PublicFeatures => new Dictionary<Type, object>();
```
is shorthand for
```
public IReadOnlyDictionary<Type, object> PublicFeatures
        {
            get
            {
                return new Dictionary<Type, object>();
            }
        }
```
New dictionary is initialized on each property call.

While 
```
public IReadOnlyDictionary<Type, object> PublicFeatures { get; } = new Dictionary<Type, object>();
```
is shorthand for
```
private readonly IReadOnlyDictionary<Type, object> _publicFeatures = new Dictionary<Type, object>();
        public IReadOnlyDictionary<Type, object> PublicFeatures => _publicFeatures;
```
where only one dictionary is initialized.

jinek:
@tomlm Why did you decide to close this one?

src/Consolonia.Core/Drawing/ConsoloniaPlatformRenderInterfaceContext.cs
tomlm: got it, dang, that's subtle and dangerous. I'm going to have be a lot more careful using that pattern.

jinek: Interesting thing that CodeRabbit did not notice this.

tomlm:
I had a typo in the branch name and renaming the branch closed the PR!

PR #159, opened 2024-11-30
Upgrade Avalonia to 11.2.1

tomlm:
Everything seems to work!


tomlm:
Do you think we should disable flyout test? I can't seem to figure out where Flyout canceling is hooked up to keyboard at all.  My only theory is that we are somehow failing to set the control focus to the flyout....Any idea?

jinek:
> Do you think we should disable flyout test? I can't seem to figure out where Flyout canceling is hooked up to keyboard at all. My only theory is that we are somehow failing to set the control focus to the flyout....Any idea?

My only quick idea is to open old version and put a breakpoint into closing or detaching from the visual tree. Thus stack would be visible.
I don't have much time, is it high priority? I'm still stuck on `theme` upgrade.

tomlm:
I'll take another crack at it.

jinek:
@tomlm I have setup CodeRabbit to review draft PRs

tomlm:
Hmmm....
On Avalonia 11.0.9 this line returns an object
```IInputElement current = TopLevel.GetTopLevel((Visual)sender)!.FocusManager!.GetFocusedElement();```
and on Avalonia 11.2.1 it returns null.

**Comment by jinek** [comment]
> **Actionable comments posted: 1**
> 
> <details>
> <summary>🧹 Outside diff range and nitpick comments (2)</summary><blockquote>
> 
> <details>
> <summary>src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs (2)</summary><blockquote>
> 
> `134-138`: **Add error handling for popup closing**
> 
> The popup closing operation should include error handling to gracefully handle any failures.
> 
> ```diff
> -if (popup != null)
> -    popup.Close();
> +if (popup != null)
> +{
> +    try
> +    {
> +        popup.Close();
> +    }
> +    catch (Exception ex)
> +    {
> +        // Log the error but don't rethrow as this is UI cleanup
> +        Debug.WriteLine($"Error closing popup: {ex.Message}");
> +    }
> +}
> ```
> 
> ---
> 
> `130-130`: **Improve code comment clarity**
> 
> The comment could be more descriptive about the relationship between overlays and popups.
> 
> ```diff
> -// if there is a overlay popup, close it
> +// Close any overlay popup (like Flyouts) when Escape is pressed
> ```
> 
> </blockquote></details>
> 
> </blockquote></details>
> 
> <details>
> <summary>📜 Review details</summary>
> 
> **Configuration used: CodeRabbit UI**
> **Review profile: CHILL**
> 
> <details>
> <summary>📥 Commits</summary>
> 
> Reviewing files that changed from the base of the PR and between d3f9ce746b8ee14df00579f7b0f91a9e43c743f6 and 32628a3439558dbde74f965dd543de59d4926be8.
> 
> </details>
> 
> <details>
> <summary>📒 Files selected for processing (1)</summary>
> 
> * `src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs` (3 hunks)
> 
> </details>
> 
> <details>
> <summary>🔇 Additional comments (2)</summary>
> 
> <details>
> <summary>src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs (2)</summary>
> 
> `19-21`: **Consider maintaining dependency injection pattern**
> 
> The constructor change removes dependency injection capability by directly instantiating `KeyboardNavigationHandler`. This creates tight coupling and makes the class harder to test.
> 
> ---
> 
> `128-141`: **Verify if this implementation correctly handles all Flyout scenarios**
> 
> The current implementation might not handle all Flyout dismissal cases, as mentioned in the PR description about Flyouts not dismissing with Escape key. Consider:
> 1. Are there cases where the Flyout is not wrapped in an OverlayPopupHost?
> 2. Should we handle error cases during popup closing?
> 3. Is focus management needed for proper Flyout handling?
> 
> <details>
> <summary>✅ Verification successful</summary>
> 
> Based on the search results, I can now provide a conclusive response about the Flyout handling implementation:
> 
> **Flyout dismissal with Escape key is working as expected**
> 
> The implementation correctly handles Flyout scenarios as evidenced by:
> 
> 1. The test file `FlyoutTests.cs` explicitly verifies that Flyouts are properly dismissed with the Escape key
> 2. Both MenuFlyout and basic Flyout implementations use the standard Avalonia popup infrastructure (OverlayPopupHost)
> 3. No competing Escape key handlers were found that could interfere with Flyout dismissal
> 
> The current implementation in `ArrowsAndKeyboardNavigationHandler.cs` is sufficient because:
> - Flyouts are always wrapped in OverlayPopupHost (as shown in the test scenarios)
> - The popup closing mechanism is simple and doesn't require additional error handling
> - Focus management is handled automatically by the popup infrastructure
> 
> 
> 
> </details>
> 
> <details>
> <summary>🏁 Scripts executed</summary>
> 
> The following scripts were executed for the analysis:
> 
> Script:
> 
> ```shell
> #!/bin/bash
> # Search for Flyout usage patterns in the codebase
> echo "Searching for Flyout usage patterns..."
> rg -A 5 "Flyout" 
> 
> # Search for other escape key handlers that might interfere
> echo "Searching for other escape key handlers..."
> rg -A 5 "Key.Escape"
> 
> # Search for popup closing patterns
> echo "Searching for popup closing patterns..."
> ast-grep --pattern 'popup.Close()'
> 
> ```
> 
> Length of output: 12417
> 
> </details>
> 
> </details>
> 
> </details>
> 
> </details>
> 
> 

I dont like coderabbit proposes such things. I had a look on setup, there is a possibility to inject instructions. I think I will aggregate several issues and will try to set the instructions.

File: src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs [hunk 1]
```diff
@@ -15,9 +16,9 @@ public class ArrowsAndKeyboardNavigationHandler : IKeyboardNavigationHandler
         //todo: check XTFocus https://github.com/jinek/Consolonia/issues/105#issuecomment-2089015880
         private IInputRoot _owner;
 
-        public ArrowsAndKeyboardNavigationHandler(IKeyboardNavigationHandler keyboardNavigationHandler)
+        public ArrowsAndKeyboardNavigationHandler()
         {
-            _keyboardNavigationHandler = keyboardNavigationHandler;
+            _keyboardNavigationHandler = new KeyboardNavigationHandler();
```
goe5mu, line 21: Could you please explain why is it null? May be in new version we need to register the handler somewhere? I mean may be in older version it was registered by something else which is not in the game anymore?

src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs
```diff
@@ -124,8 +125,26 @@ private void OnKeyDown(object sender, KeyEventArgs e)
         {
             if (e.Handled) return;
 
+            if (e.Source is MenuItem) return;
```
goe5mu, line 128: I'm wondering how does Avalonia work without this? May be we should drop a ticket to Avalonia same time?

goe5mu (approved):

src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs
tomlm: The old code base used to instantiate a KeyboardHandler in the ConsoloniaApplication and then pass it in to our registered instance as a singleton. So both the keyboard handler and the ArrowsAndKeyboard handler were singletons.

The new Avalonia code base has KeyboardHandler as a transient in the AvaloniaLocator, which means it is created on demand. According to documentation and discussions AvaloniaLocator is an internal resource resolver and not meant for general purpose dependency resolution.  We are attempting to replace the one that is registered with the one that we have which wraps it. Soo...I looked at the registration of the dependency in the Avalonia side of things and it's just newing up a new instance, making it transient.  I was not able to create a new registration in AvaloniaLocator for the outer navigator without creating a recursive dependency, so I elected to just instantiate it in the class.

If you have a cleaner way of doing that I'm open to it.

src/Consolonia.Core/Infrastructure/ArrowsAndKeyboardNavigationHandler.cs
tomlm: I don't really understand why we have the ArrowsAndKeyboardNavigatorHandler at all. I was thinking about this and probably the cleaner thing to do would be to register the OnKeyDown hook if the item is a MenuItem. 

I can give that a try if you think it's cleaner...

jinek: > All I know is that if there is a MenuItem then the left/right/esc keys navigation keys logic messes up the keyboard navigation when in a menu system.

ok. It looks like a bug of Avalonia. MenuItem should set e.Handled if it handles arrow navigation. I will investigate after.

jinek (approved):

PR #160, opened 2024-11-30
fix issue 158 Shading is computed wrong.

tomlm:
* added unit tests

image


src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
```diff
@@ -125,9 +125,14 @@ public Pixel Blend(Pixel pixelAbove)
...
                 case PixelBackgroundMode.Shaded:
+#if DEBUG
+                    if (pixelAbove.Foreground.Symbol.Text != " ")
+                        throw new ArgumentOutOfRangeException(nameof(pixelAbove),
+                            "Someone is attempting to shade a pixel with a non-blank symbol and the behavior of that is not defined");
+#endif
                     (newForeground, newBackground) = Shade();
-                    newForeground = newForeground.Blend(pixelAbove.Foreground);
-                    break;
+                    return new Pixel(newForeground, newBackground);
```
jinek, line 134: Lets discuss. We can still draw a non blank symbol on top of the shade, right?

tomlm: Yes. Am I losing something by converting the shade background to a transform on the existing color at this point?

jinek: Almost
> Do you think I should do:
Shaded => Dim() existing current pixel background, then blend current foreground, then I suppose Dim() the result?

I'm thinking we should dim the foreground first, then just blend with foreground of new pixel. 
This way, if there is no foreground above then foreground of old pixel is just dimmed. But if there is a new pixel, then it should be preserved and not shaded.

jinek: I just understood, that before it was probably fine. But now we have different meaning of `Blend` procedure. I think it's correct you are getting rid of it here, but we just need to put correct symbol on top if there is one.

jinek (approved):

PR #175, opened 2024-12-05
Prepare build and packages for publishing

tomlm:
This is to implement #169  and #172 
* Added Icon.png to each package
* Added Readme.txt to each package
* Consolidated GuiCS into Consolonia.Platform package so there isn't a standalone GuiCS package
* Added github actions for publishing packages to nuget


.github/workflows/PublishIncrementalNuget.yml
```diff
@@ -0,0 +1,49 @@
...
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 
+        uses: actions/checkout@v4
+  
+      - name: Setup .NET
```
jinek, line 13: Can we extract repeated things to somewhere? Would be nice to re-use same things in different workflows

src/Consolonia.Core/Consolonia.Core.csproj
```diff
@@ -12,4 +12,9 @@
         <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     </ItemGroup>
 
+    <ItemGroup>
+        <None Include="../../Icon.png" Pack="True" PackagePath="\" />
```
jinek, line 16: We can setup it in Directory.Build.props not to mention in each project

jinek:
Just small clarification. Directory.Build.props is automatically included to all projects. Directory.Core.Build.props must be included manually to projects where needed.

src/Consolonia.GuiCS/Consolonia.GuiCS.csproj
```diff
@@ -1,7 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Core.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
jinek, line 2: What do we want to apply from there?

src/Tools/Avalonia.PreviewHost/Avalonia.PreviewHost.csproj
```diff
@@ -5,8 +5,9 @@
         <Nullable>enable</Nullable>
         <BuiltInComInteropSupport>true</BuiltInComInteropSupport>
         <ApplicationManifest>app.manifest</ApplicationManifest>
+        <IsPackable>false</IsPackable>
```
jinek, line 8: May be we should re-arrange build.props if we have more than two setup for assemblies

foo.cmd
```diff
@@ -0,0 +1,2 @@
+erase /q packages\*.*
+dotnet pack src/Consolonia.sln -c Release -o packages --include-symbols --property WarningLevel=0
```
jinek, line 2: is this file needed?

.github/workflows/PublishIncrementalNuget.yml
tomlm: I have no idea. I created these a long time ago and I just keep copy-pasta'ing them everywhere because they work! :)

src/Consolonia.Designer/README.md
```diff
@@ -0,0 +1,26 @@
+# Consolonia.Designer
+Consolonia is a TUI (Text User Interface) (GUI Framework) implementation for [Avalonia UI](https://github.com/AvaloniaUI)
```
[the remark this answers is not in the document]
tomlm: Right now the readme files for each package are in the same layout but with package specific details.

src/Consolonia.GuiCS/Consolonia.GuiCS.csproj
tomlm: this is magic build props which makes it so that the GuiCS is included as an assembly in PlatformSupport package, and doesn't live as it's own top level package.

foo.cmd
tomlm: wooops! Nope, I shouldn't have checked that in, that's my testing of the what the build does to make packages so I can inspect them. I will delete

.github/workflows/PublishIncrementalNuget.yml
tomlm: can we do this for a later checkin? I want to work more in themes area and this delta renamed everything...

src/Tools/Avalonia.PreviewHost/Avalonia.PreviewHost.csproj
tomlm: I have moved Avalonia.PreviewHost and Consolonia.Designer into an Experimental folder and added comment stating that these are marked as not packable because they are still experimental.

src/Consolonia/Assembly.cs
```diff
@@ -0,0 +1,6 @@
+using System;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Consolonia.Core.Tests")]
+
+[assembly: CLSCompliant(true)] //todo: should we make it compliant?
```
jinek, line 6: Btw, why do we need to make it CLS compliant?

jinek (approved):

PR #199, opened 2024-12-10
Add 2 new line styles for brushes Edge and EdgeWide

tomlm:
image

Edge uses 1/8 wide vertical/horizontal chars
EdgeWide uses 1/4 vertical/horizontal block chars (remember the QuadPixels?)

Benefits of Edge and EdgeWide (and Bold) is that it is tight up next to the edge of content, which means that you don't end up with double background color effects like this:
image

> NOTE: Only BorderThickness=1 is supported.  

Non-uniform borders work now too (as long as you set appropriate padding.
image

Related Avalonia bug https://github.com/AvaloniaUI/Avalonia/issues/17752 filed


tomlm:
* Added 2 new linestyles **Edge** and **EdgeWide**

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -126,6 +149,86 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
                 case Line myLine:
                     DrawLineInternal(pen, myLine);
                     break;
+                case StreamGeometryImpl streamGeometry:
+                {
+                    pen = pen ?? new Pen(brush);
+
+                    var extractColorCheckPlatformSupported =
+                        ExtractColorOrNullWithPlatformCheck(pen, out var lineStyle);
+                    if (extractColorCheckPlatformSupported == null)
```
jinek, line 158: Does it make sense to ignore no supported? Could be better to notify it's not going to be painted with an exception?

src/Consolonia.Core/Drawing/StreamGeometryImpl.cs
```diff
@@ -0,0 +1,197 @@
...
+        public double ContourLength => _strokes.Sum(l => l.ContourLength);
+
+
+        public IStreamGeometryImpl Clone()
+        {
+            var clone = new StreamGeometryImpl();
+            foreach (Line line in _strokes) clone._strokes.Add(line);
+            foreach (Rectangle rect in _fills) clone._fills.Add(rect);
+            clone._bounds = _bounds;
+            return clone;
```
jinek, line 39: Half-deep cloning?) I'm not aware how it's used, but is it fine we keep same instances of lines and rectangles for the clone? Should we do serializing/deserializing or other deep-cloning techniques?

src/Consolonia.Core/Drawing/StreamGeometryImpl.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        public bool FillContains(Point point)
```
jinek, line 42: Should not this exclude the border?

jinek: like I mean:   return _bounds.Contains(point) && !StrokeContains(point)

jinek (approved):

jinek:
@tomlm I left questions mostly and disabled auto-merge. Feel free to merge please.

src/Consolonia.Core/Drawing/StreamGeometryImpl.cs
tomlm: I updated this to clone the classes, all of the parts that "copied" now are structs and immutable.

src/Consolonia.Core/Drawing/StreamGeometryImpl.cs
tomlm: yeah, I updated...

Issue #196, opened 2024-12-08
Alt Key accelerators aren't rendered


tomlm:
Menu, Button, etc.

I looked into it and figured this much out:
* There is a **AccessKeyHandler** which listens at the visual root level. 
* Whenever a control is part of the parent it registers (somehow) with it to be notified when an AccessKey should be displayed. 
* The **AccessKeyHandler** has an owner set to it, and when it has an owner it registers for key events and toggles the ShowAccessKeys property, which triggers all of the children to render with the _ on the character.
* Each element looks for the AccessKeyHandler (somehow) and subscribes to changes on the ShowAccessKeys property.
* ConsoleWindow was not mapping Alt key correctly, it was coming through as a space.

I updated ConsoleWindow.cs to look like this:
```c#
 private void HandleKeyInput(WindowsConsole.InputRecord inputRecord)
 {
     WindowsConsole.KeyEventRecord keyEvent = inputRecord.KeyEvent;
     RawInputModifiers modifiers =
         ModifiersFlagTranslator.Translate(keyEvent.dwControlKeyState);
     char character = keyEvent.UnicodeChar;
     Avalonia.Input.Key key;
     if (character == '\0')
     {
        // modifier only key, there is no character, avalonia maps this as a Key.LeftAlt key etc.
         switch (modifiers)
         {
             case RawInputModifiers.Control:
                 key = Avalonia.Input.Key.LeftCtrl;
                 break;
             case RawInputModifiers.Alt:
                 key = Avalonia.Input.Key.LeftAlt;
                 break;
             case RawInputModifiers.Shift:
                 key = Avalonia.Input.Key.LeftShift;
                 break;
             default:
                 key = Avalonia.Input.Key.None;
                 break;
         }
     }
     else
         key = DefaultNetConsole.ConvertToKey((ConsoleKey)keyEvent.wVirtualKeyCode);

     RaiseKeyPress(key, character, modifiers, keyEvent.bKeyDown, (ulong)Stopwatch.GetTimestamp());
 }
```
Much to my dismay, this did not fix it. I **think** that somehow the accesskeyhandler is not being registered correctly because our root window is a ConsoleWindow, but I haven't figured out where/how that happens.  It looks like it might have something to do with TopLevel, which we don't even implement?  Each control calls RegisterServices and inside there it seems to create an accesshandler...?

Anyhow. I though I would file this bug with my learnings to hopefully help either me or the next guy who comes along and tries to fix this.

jinek:
[...]

PR #205, opened 2024-12-14
Add ConsoloniaAccessText underline

tomlm:
Fix for #196 
accessText

### DIFF
--- a/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
+++ b/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
@@ -1,5 +1,4 @@
 using System;
-using Avalonia.Media;
 using Avalonia.Controls.Primitives;
 using Avalonia.Controls.Documents;
 
@@ -26,21 +25,18 @@ namespace Consolonia.Core.Controls
                     {
                         if (!String.IsNullOrEmpty(Text))
                         {
-                            _accessRun = null;
                             InlineCollection inlines = new InlineCollection();
                             var iPos = Text.IndexOf('_', StringComparison.Ordinal);
                             if (iPos >= 0 && iPos < Text.Length - 1)  
                             {
-                                inlines.Add(new Run(Text.Substring(0, iPos)));
-                                _accessRun = new Run(Text.Substring(++iPos, 1))
-                                {
-                                    TextDecorations = new TextDecorationCollection()
-                                };
+                                inlines.Add(new Run(Text[..iPos]));
+                                _accessRun = new Run(Text[++iPos..++iPos]);
                                 inlines.Add(_accessRun);
-                                inlines.Add(new Run(Text.Substring(iPos + 1)));
+                                inlines.Add(new Run(Text[iPos..]));
                             }
                             else
                             {
+                                _accessRun = null;
                                 inlines.Add(new Run(Text));
                             }
                             this.Inlines = inlines;
@@ -52,9 +48,9 @@ namespace Consolonia.Core.Controls
                     if (_accessRun != null)
                     {
                         if (this.ShowAccessKey)
-                            _accessRun.TextDecorations.Add(new TextDecoration { Location = TextDecorationLocation.Underline });
+                            _accessRun.TextDecorations = Avalonia.Media.TextDecorations.Underline;
                         else
-                            _accessRun.TextDecorations.Clear();
+                            _accessRun.TextDecorations = null;
                     }
                     break;
             }
--- a/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
+++ b/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
@@ -9,20 +9,13 @@ namespace Consolonia.Core.Controls
     /// <summary>
     /// A text block that displays a character prefixed with an underscore as an access key.
     /// </summary>
-    public class ConsoloniaAccessText : AccessText, IDisposable
+    public sealed class ConsoloniaAccessText : AccessText
     {
-        private bool _disposed;
         private Run _accessRun;
 
         public ConsoloniaAccessText()
         {
             this.PropertyChanged += OnPropertyChanged;
-        }
-
-        public void Dispose()
-        {
-            this.PropertyChanged -= OnPropertyChanged;
-            _disposed = true;
         }
 
         private void OnPropertyChanged(object sender, Avalonia.AvaloniaPropertyChangedEventArgs e)
@@ -63,9 +56,6 @@ namespace Consolonia.Core.Controls
                         else
                             _accessRun.TextDecorations.Clear();
                     }
-                    break;
-
-                default:
                     break;
             }
         }
--- a/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
+++ b/src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
@@ -1,12 +1,12 @@
 using System;
-using Avalonia.Controls.Primitives;
+using Avalonia;
 using Avalonia.Controls.Documents;
+using Avalonia.Controls.Primitives;
 
 namespace Consolonia.Core.Controls
 {
-
     /// <summary>
-    /// A text block that displays a character prefixed with an underscore as an access key.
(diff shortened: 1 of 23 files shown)


src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
```diff
@@ -0,0 +1,63 @@
...
+            switch (e.Property.Name)
+            {
+                case nameof(Text):
+                    {
+                        if (!String.IsNullOrEmpty(Text))
+                        {
+                            _accessRun = null;
+                            InlineCollection inlines = new InlineCollection();
+                            var iPos = Text.IndexOf('_', StringComparison.Ordinal);
+                            if (iPos >= 0 && iPos < Text.Length - 1)
```
jinek, line 32: I think complex condition can be replaced with just iPos!=-1

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
```diff
@@ -0,0 +1,63 @@
...
+                case nameof(Text):
+                    {
+                        if (!String.IsNullOrEmpty(Text))
+                        {
+                            _accessRun = null;
+                            InlineCollection inlines = new InlineCollection();
+                            var iPos = Text.IndexOf('_', StringComparison.Ordinal);
+                            if (iPos >= 0 && iPos < Text.Length - 1)  
+                            {
+                                inlines.Add(new Run(Text.Substring(0, iPos)));
```
jinek, line 34: Just a hint: Text[..iPos] or Text[iPos..]

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        private void OnPropertyChanged(object sender, Avalonia.AvaloniaPropertyChangedEventArgs e)
+        {
+            switch (e.Property.Name)
+            {
+                case nameof(Text):
+                    {
+                        if (!String.IsNullOrEmpty(Text))
+                        {
+                            _accessRun = null;
```
jinek, line 29: Looks not necessary

jinek (commented):
Button and other controls use some kind of factory or template, which can be also overridden to use ConsoloniaAccessText.

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
```diff
@@ -0,0 +1,66 @@
...
+                                _accessRun.TextDecorations.Add(new TextDecoration { Location = TextDecorationLocation.Underline });
+                            else
+                                _accessRun.TextDecorations.Clear();
+                        }
+                        break;
+                    default:
+                        break;
+                }
+
+            };
```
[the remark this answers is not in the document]
jinek: How can the memory leak be caused if the event is instance, not static?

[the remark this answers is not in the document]
jinek: @tomlm do we use it? Thinking again to switch it off. It sometimes seems we are heating earth for nothing.

tomlm:
I was looking, couldn't find it. Any idea where?

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
tomlm: oh, fancy. I totally forgot about that new feature. It hasn't been integrated into my coding patterns.

tomlm: Holy shit, I just tried it...so clean, so readable, so succinct. .Substring always felt so verbose. I love it!  Thanks for the tip!

tomlm:
it works almost.
There are 2 issues left:
**Accelerator used for Label puts focus on the input control AND puts accelerator key into the input**
label
Alt+F puts focus on the target element, but then even though I set the e.Handled=true it puts the letter 'F' into the input control. 
It should just set the focus and not add the 'F'.

**Toggle of Alt accelerator works only after the FIRST time I use Menu**
This is super weird. 
On Avalonia if you hit alt key, it shows the underscore, then hit Alt key again it hides it. If you hit the accelerator, it activates the target and hides it.
In Consolonia, if you hit the Alt key it shows the underscore, then hit Alt key again it leaves it shown. 
Unless I go to the Menu page. When I go to the menu page, then it starts working, and it works until I shut down the app. 
WTF???

label2

tomlm:
The second is an avalonia bug. I just repoed in avalonia proper.

jinek:
First one I can check.
Do I need to review now? Lets please use draft PRs to indicate it's not for review in case it's not. It's sometimes hard for me to prioritize what to do here.

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
```diff
@@ -0,0 +1,66 @@
...
+                case nameof(Text):
+                {
+                    if (!string.IsNullOrEmpty(Text))
+                    {
+                        var inlines = new InlineCollection();
+                        int iPos = Text.IndexOf('_', StringComparison.Ordinal);
+                        if (iPos >= 0 && iPos < Text.Length - 1)
+                        {
+                            inlines.Add(new Run(Text[..iPos]));
+                            _accessRun = new Run(Text[++iPos..++iPos]);
```
jinek, line 38: it looks stunning!)

jinek:
> it works almost. There are 2 issues left: **Accelerator used for Label puts focus on the input control AND puts accelerator key into the input**

Could you please push the repro of label+input somewhere?

tomlm:
It should be pushed....there should be a panel page in gallery

tomlm:
I'm ready to check it in. Sorry. I was ready, then based on feedback on avalonia side I made some changes, and then feedback from you I added the Label/Button support. etc. 

I will try to be more disclipined about using Draft PR. Sorry.

src/Consolonia.Core/Controls/ConsoloniaAccessText.cs
tomlm: 🤪

jinek:
> I'm ready to check it in. Sorry. I was ready, then based on feedback on avalonia side I made some changes, and then feedback from you I added the Label/Button support. etc.
> 
> I will try to be more disclipined about using Draft PR. Sorry.

no worries. But I want to investigate ALT handling also, the guy you've changed.

Thought I have investigated the issue of producing excessive F.
Issue happens because `ConsoleWindow.ConsoleOnKeyEvent` produces `RawTextInputEventArgs` after the focus was already set because previously `RawKeyEventArgs.Handled` was not set to true by AccessKey infrastructure.
It looks like either `Dispatcher.UIThread.InvokeAsync` does not wait for the delegate to complete or may be this logic has been changed completely. I will check if Avalonia has reworked this part, we may be just need to reflect those changes from new version of Avalonia.

tomlm:
I bet we should be awaiting the InvokeAsync. There are 2 patterns, there is Post() and Invoke(). Invoke returns a task, we should be careful about the semantics of which one we are using.

jinek:
> I bet we should be awaiting the InvokeAsync. There are 2 patterns, there is Post() and Invoke(). Invoke returns a task, we should be careful about the semantics of which one we are using.

This one excluded. Task returned from Invoke properly awaits for access key to be processed.

I could not find matching code in Avalonia at all. Thinking may be I just invented this processing: if KeyDown event was not handled, then producing TextInput event.
I will try to do the trick by checking if a key was pressed together with ALT.

jinek:
Yeah, that helped. fixed first issue

tomlm:
Second issue is Avalonia bug. It looks like we can work around by having an empty MENU on the page.

jinek:
@tomlm 
please, check my last commit. I have removed special handling of ALT key. It seems all works fine without

tomlm:
Looks good.

jinek (approved):

tomlm:
Thanks Evgeny! 

If you are ever in Seattle I'll buy you a beer!

jinek:
No, I will buy you a beer. 😃

Issue #179, opened 2024-12-05
Caret is not drawn by CaretControl


jinek:
I'm not sure whether it's intentionally or no, most of the controls contain CaretControl which is supposed to draw caret when the control is focused.
As an example, on the screenshot item "Button" within the grid is currently focused, but there is no caret there:
![image](https://github.com/user-attachments/assets/19db05b0-ea34-4e52-9b6c-cf0f7deb8b87)

tomlm:
what does it mean to draw caret? Is it turn on blinking cursor? Or something else?

jinek:
Yes. Before smth it was blinking over focused element

tomlm:
I can take a look at this.

tomlm:
Hmm...I thought maybe this was something I broke, but the current logic only seems to apply a caret inside of TextBox and it works correctly there.

The current template for CaretControl doesn't seem to ever activate the IsPixel logic. I don't think understand how this works (and how it's supposed to work)

PR #212, opened 2024-12-16
fix caret cursor

tomlm:
#179 Caret is not being drawn by caret control.
For some reason the Line template wasn't being drawn. It never called into DrawingContext. I think it may be because we were drawing a line which had Width and Height but no start point or end point. I verified that Avalonia GUI ignores Lines with just width/height, so this is avalonia behavior.
I changed CaretControl template to output a Rectangle of width/1,height/1 and now it calls, gets rendered correctly.

### DIFF
--- a/src/Consolonia.Core/Drawing/CaretStyle.cs
+++ b/src/Consolonia.Core/Drawing/CaretStyle.cs
@@ -1,6 +1,6 @@
 ﻿namespace Consolonia.Core.Drawing
 {
-    public enum CursorStyle
+    public enum CaretStyle
     {
         BlinkingBar,
         SteadyBar,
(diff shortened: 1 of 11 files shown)


goe5mu:
Caret does not show up in ComboBoxItem. Visible when opening ComboBox.

jinek:
Caret does not show up in ComboBoxItem. Visible when opening ComboBox.

tomlm:
ComboBoxItem behavior is baked into Avalonia.
```csharp
   public ComboBoxItem()
   {
       this.GetObservable(ComboBoxItem.IsFocusedProperty)
           .Subscribe(focused =>
           {
               if (focused)
               {
                   (Parent as ComboBox)?.ItemFocused(this);
               }
           });
   }
```

tomlm:
Hmmm....I can't seem to figure out how to get comboboxItem to show focus.

tomlm:
I figured it out. I need to draw the movebrush AFTER the content is drawn, not before, otherwise the content resets the MoveBrush pixel

Issue #138, opened 2024-11-03
ScrollViewer doesn't scroll if cursor is not over a child element


tomlm:
# Repo
go to Image gallery

* Put mouse over the images and use mouse wheel => it scrolls
* Put mouse over background of scroll window and use mouse wheel => it doesn't scroll

# Expected
Mouse wheel should scroll inside the entire rectangle of the scroll area.

jinek:
I have been fixing several related issues. Had to set some background, even tho ttansparent, but not null, as I remember. It helped with datagrid and listbox scrolling.

PR #214, opened 2024-12-17
Enable scrolling behavior to work correctly by making sure there is background of transparent on the panels for containers.

tomlm:
* Update Scroller to have bigger scroll area so it's clearer to see it working 
* Fixed bug in AccessText registration which was breaking comboboxes.

Bug fix for #138 ScrollViewer doesn't work if not over a child element.


src/Consolonia.Gallery/Gallery/GalleryViews/GalleryScrollViewer.axaml.cs
```diff
@@ -61,6 +89,12 @@ public ScrollBarVisibility VerticalScrollVisibility
         }
 
         public List<ScrollBarVisibility> AvailableVisibility { get; }
+
+        public List<string> Cows
```
jinek, line 93: I'm thinking that is miss we are using lists rather than arrays. To support UI update then we could use OvservableCollection or BindingCollection or whatever, but List is worst option

jinek (commented):
I'm proposing not to bring one more copy or RX extension, or do we need it in the future?

src/Consolonia.Themes/Templates/Controls/ListBoxItem.axaml
```diff
@@ -10,10 +10,11 @@
                 Value="{DynamicResource ThemeTransparentBrush}" />
         <Setter Property="Template">
             <ControlTemplate>
-                <Panel Background="{TemplateBinding Background}">
+                <Panel Background="Transparent">
```
jinek, line 13: Do we know why ThemeTransparentBrush did not do the job?

src/Consolonia.Core/Helpers/Observable.cs
```diff
@@ -0,0 +1,119 @@
+using System;
+using Avalonia.Reactive;
+
+namespace Consolonia.Core.Helpers
+{
+    /// <summary>
+    ///     Provides common observable methods as a replacement for the Rx framework.
+    /// </summary>
+    /// <remarks>
+    ///     COPIED FROM AVALONIA.BASE BECAUSE IT IS NOT PUBLICLY ACCESSIBLE,
```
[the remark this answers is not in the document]
tomlm: agreed, I wasn't including it because this is what Avalonia was doing. I just couldn't figure out how to do the binding like you showed me, that removes the need for this entirely. Thanks,

src/Consolonia.Themes/Templates/Controls/ListBoxItem.axaml
tomlm: no.

tomlm: huh. I think the scrollviewer fix makes the listbox work without any changes.

tomlm: I don't know why that theme brush didn't work. When I use hard-coded "Transparent" it works.

jinek (approved):

PR #219, opened 2024-12-19
203-selectabletextblock-does-not-work

jinek:


src/Consolonia.Themes/Templates/Controls/Helpers/SelectTextWithPointerUpExtension.cs
```diff
@@ -0,0 +1,57 @@
...
+            AvaloniaProperty.RegisterAttached<SelectableTextBlock, bool>(CommonInternalHelper.GetStyledPropertyName(),
+                typeof(SelectTextWithPointerUpExtension));
+
+        static SelectTextWithPointerUpExtension()
+        {
+            var console = AvaloniaLocator.Current.GetService<IConsole>();
+            bool supportsMouse = console.SupportsMouse;
+            bool supportsMoveMove = console.SupportsMouseMove;
+            if (!supportsMouse || supportsMoveMove)
+                return;
```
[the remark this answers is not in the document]
jinek: If mouseMove is supported then SelectableTextBlock behaves itself normally, we don't need this subscription.

src/Consolonia.Themes/Templates/Controls/SelectableTextBlock.axaml
```diff
@@ -0,0 +1,11 @@
+<ResourceDictionary xmlns="https://github.com/avaloniaui"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:helpers="clr-namespace:Consolonia.Themes.Templates.Controls.Helpers">
+    <ControlTheme x:Key="{x:Type SelectableTextBlock}"
+                  TargetType="SelectableTextBlock">
+        <Setter Property="SelectionBrush"
+                Value="{DynamicResource ThemeActionBackgroundBrush}" />
+        <Setter Property="helpers:SelectTextWithPointerUpExtension.SelectOnMouseLeftUp"
```
tomlm, line 8: next level magic. 

I haven't ever created a mixin extension like this.  Xaml is so freaking powerful as a model.

jinek: True. Its supposed to be simpler with "behaviours", they manage lifetime etc, but Avalonia does not have them builtin, i dont want to include that assembly to core level packages of consolonia.

PR #224, opened 2024-12-23
Add new Consolonia.Templates template package

tomlm:
Adds new cli template
```
> dotnet new install Consolonia.Templates
> dotnet new list consolonia
These templates matched your input: 'consolonia'

Template Name           Short Name  Language  Tags
----------------------  ----------  --------  -------
Consolonia Application  consolonia  [C#]      console
> dotnet new consolonia
> dotnet run
```
Will output a consolonia project with app and mainwindow defined.


src/Consolonia.Gallery/Gallery/GalleryViews/GalleryScrollViewer.axaml.cs
```diff
@@ -66,7 +66,7 @@ public ScrollViewerPageViewModel()
                                       `^'     \ :/           `^'  `-^-'   \v/ :  \/ 
                               Bill Ames
 
-                              """.Split(['\r', '\n'], StringSplitOptions.RemoveEmptyEntries));
+                              """.Split((char[])['\r', '\n'], StringSplitOptions.RemoveEmptyEntries));
```
jinek, line 69: I'm curios what was the type before cast

tomlm: super weird. I'm using my ARM tablet and build is broken on ARM because it is ambigious between 
```
.Split(char[], ...)
.Split(string, ...)
```
methods. 
I think this is a dotnet bug in the arm implementation or something.  I couldn't build without explicitely casting.

tomlm: I think string has char[] cast operator or something.

jinek: very interesting. ARM64? what is the device?

jinek (approved):

Issue #190, opened 2024-12-05
Strange coloring in TurboVision theme


jinek:
Background seems darker now (on the left), border became white, green is different.
![image](https://github.com/user-attachments/assets/30e7308e-3214-4170-98f6-6695a33813ed)

Other colors are somehow also different:
![image](https://github.com/user-attachments/assets/48aef156-089b-4534-ba95-f572bb8c68aa)
![image](https://github.com/user-attachments/assets/97731539-365f-4e45-b18b-c691a4e64227)

Other pages as well. It looks like all this have one cause.

jinek:
https://github.com/jinek/Consolonia/discussions/166

PR #227, opened 2024-12-25
190-strange-coloring-remove-console-brush

jinek:


jinek:
Should not close this one https://github.com/jinek/Consolonia/issues/190

src/Consolonia.Core/Drawing/ConsoleBrush.cs
```diff
@@ -160,21 +80,18 @@ public static ConsoleBrush FromPosition(IBrush brush, int x, int y, int width, i
...
                 }
 
+                case ISolidColorBrush solidColorBrush:
+                    return solidColorBrush.Color;
+
                 default:
-                    return FromBrush(brush);
+                    ConsoloniaPlatform.RaiseNotSupported(751, brush); //todo: allow RaiseNotSupported to return a result
+                    return new Color();
             }
```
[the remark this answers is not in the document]
jinek: RaiseNotSupported will throw an exception by default

src/Consolonia.Core/Controls/Dialog/DialogWrap.axaml
```diff
@@ -1,10 +1,9 @@
 <UserControl xmlns="https://github.com/avaloniaui"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:drawing1="clr-namespace:Consolonia.Core.Drawing;assembly=Consolonia.Core"
              x:Class="Consolonia.Core.Controls.Dialog.DialogWrap">
 
     <Panel>
-        <Border Background="{drawing1:ConsoleBrush Mode=Shaded}"
+        <Border Background="{DynamicResource ThemeShadeBrush}"
```
tomlm, line 6: oooh exciting

src/Consolonia.Core/Drawing/ConsoleBrush.cs
tomlm, line 1: rename file?

src/Consolonia.Core/Drawing/ConsoleBrush.cs
```diff
@@ -1,96 +1,14 @@
...
-                    return new ConsoleBrush(solidBrush.Color, mode ?? PixelBackgroundMode.Colored);
-
-                default:
-                    ConsoloniaPlatform.RaiseNotSupported(6);
-                    return null;
-            }
-        }
-
-        public static ConsoleBrush FromPosition(IBrush brush, int x, int y, int width, int height)
+        public static Color FromPosition(IBrush brush, int x, int y, int width, int height)
```
tomlm, line 9: do you think using 'this' would be bad?  Or do you prefer them as static methods? 
brush.FromPosition(x,y,width, height)
vs
BrushExtensions.FromPosition(brush, x,y,heigh)?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -731,14 +721,13 @@ private void DrawLineSymbolAndMoveHead(ref Point head, bool isVertical, ISymbol
...
-            foreground = ConsoleBrush.FromBrush(foreground);
-            if (foreground is not ConsoleBrush { Mode: PixelBackgroundMode.Colored } consoleBrush)
+            if (foreground is not ISolidColorBrush solidColorBrush)
             {
                 ConsoloniaPlatform.RaiseNotSupported(4);
                 return;
             }
 
-            // if (!Transform.IsTranslateOnly()) ConsoloniaPlatform.RaiseNotSupported(15);
+            // if (!Transform.IsTranslateOnly()) ConsoloniaPlatform.RaiseNotSupported(15); //todo: what to do if a rotation?
```
tomlm, line 730: we could theoritcally handle 90, 180, 270, 360 rotations and flip x or y transforms.

src/Consolonia.Core/Drawing/PixelBufferImplementation/DrawingBoxSymbol.cs
```diff
@@ -31,16 +31,16 @@ public bool Equals(DrawingBoxSymbol other)
 
         [JsonIgnore] public string Text { get; init; }
 
-        [JsonIgnore] public ushort Width { get; } = 1;
+        [JsonIgnore] public ushort Width => 1;
 
-        public bool IsWhiteSpace()
+        public bool NothingToDraw()
```
tomlm, line 36: much better name.

tomlm: or perhaps IsEmpty?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
```diff
@@ -102,56 +102,43 @@ public bool Equals(Pixel other)
...
-                    // blend the pixelAbove foreground into the shaded pixel
-                    newForeground = newForeground.Blend(pixelAbove.Foreground);
-
-                    // resulting in new pixel with shaded background and blended foreground
-                    return new Pixel(newForeground, newBackground);
-
-                default: throw new ArgumentOutOfRangeException(nameof(pixelAbove));
+                }
+                else
+                {
```
tomlm, line 124: add a comment to make it clear the rest of the arguments are passed through
// merge the PixelForeground color with the pixelAbove backbround color

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
```diff
@@ -102,56 +102,43 @@ public bool Equals(Pixel other)
...
+                    newForeground = new PixelForeground(Foreground.Symbol,
+                        MergeColors(Foreground.Color, pixelAbove.Background.Color),
+                        Foreground.Weight,
+                        Foreground.Style,
+                        Foreground.TextDecoration);
+
+                    if (pixelAbove.Background.Color.A == 0xFF)
+                        // non-transparent layer above
```
tomlm, line 132: hmm. so a pixelabove with opaque background always replaces. Makes sense. This is essentially Solid mode...but in this case  no need to do lines 125-129. I would move 125-129 into else statement

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
```diff
@@ -102,56 +102,43 @@ public bool Equals(Pixel other)
...
+                        newForeground = new PixelForeground();
+                }
+
+                newIsCaret = pixelAbove.IsCaret;
```
tomlm, line 136: not this delta but I think we can greatly simplify the logic by pulling IsCaret off of Pixel entirely and move to property of PixelBuffer
```PixelBuffer.Caret = X/Y coordate```
We are using caret for focus, and Avalonia never has more than a single focus. And Console Apps don't (as far as I know) have the ability to have multiple carets. 

Why not save ourselves the complexity (and even memory) ? 

But we should do this on a different PR

File: src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelForeground.cs [hunk 9]
```diff
@@ -96,5 +96,10 @@ public override int GetHashCode()
         {
             return !left.Equals(right);
         }
+
+        public bool NothingToDraw()
+        {
+            return Color.A == 0x0 || Symbol.NothingToDraw();
```
tomlm, line 102: now we are talking!

src/Consolonia.Themes/Fluent/FluentColors.axaml
```diff
@@ -1,45 +1,43 @@
 <ResourceDictionary xmlns="https://github.com/avaloniaui"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-                    xmlns:drawing="clr-namespace:Consolonia.Core.Drawing;assembly=Consolonia.Core"
```
tomlm, line 3: not this delta but we sould move LineBrush into Consolonia.Controls namespace.

src/Consolonia.Themes/Templates/Controls/CalendarItem.axaml
```diff
@@ -33,7 +32,7 @@
                                 <Style Selector="Button.CalendarHeader:pressed  /template/ ContentPresenter">
                                     <!--todo: needs some color may be-->
                                     <Setter Property="Background"
-                                            Value="{drawing:ConsoleBrush Mode=Transparent}" />
+                                            Value="{DynamicResource ThemeTransparentBrush}" />
```
tomlm, line 35: do you think there is ever a need for a visible transparent brush that would be themed?  Why not just Brushes.Transparent?

tomlm (approved):
My feedback are suggestions/comments, but this looks good.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: um...too much egg nog. I forgot FONT chars would have to transform as well. LOL

tomlm (changes requested):
image
I think this used to show compound unicode chars. Shoot, I just checked main and main does the same thing.

We have a regression in unicode rendering of compound characters. <sigh>
#229 filed

tomlm (dismissed):
the unicode regression is not yours. i will file a bug against myself to investigate.

src/Consolonia.Core/Drawing/PixelBufferImplementation/DrawingBoxSymbol.cs
jinek: Honestly I had intention to write a sentence with an explanation there. I bet I need just 1 day to forget whether space is empty or no)

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
jinek: fixed

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
jinek: fixed

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
jinek: I'm not yet convinced it would be better. Caret covered by a background or foreground is supposed to dissapear similar to a foreground of a foreground below. How would this look like? lets try

src/Consolonia.Core/Drawing/ConsoleBrush.cs
jinek: fixed

src/Consolonia.Themes/Templates/Controls/CalendarItem.axaml
jinek: My only thought here is it can be required to be defined by another type, not Brush. But probably does not matter now. I'm good to make a separate PR for this just not to mix

jinek:
> My feedback are suggestions/comments, but this looks good.

fixed several

src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
tomlm: I see. Sigh

tomlm (approved):

PR #240, opened 2024-12-30
Support switching between AnsiOutput and Console.XXX() methods 

tomlm:
This PR 
* splits **IConsole** into **IConsoleOutput**.
* adds PrepareConsole() and RestoreConsole() to initalize console and cleanup from console as appropriate (either using ANSI or Console APIs)

**IConsole** Implementations derive from **ConsoleBase** (formally known as InputLessConsole)
* **Win32Console** - uses Console Win32 API to implement mouse/keyboard support
* **CursesConsole** -- uses Curses to implement mouse/keyboard support
* **DefaultNetConsole** - Uses dotnet core classes to implement keyboard only support (no mouse)
* **_AnsiConsole_** -  (TBD) implement **AnsiConsole()** which uses Ansi escape codes for input/mouse, but this is not in this PR. It's a little bit tricky as it actually relies on the Read() functionality of Win32/Curses, etc so I'm not quite certain how to factor it yet.

**IConsoleOutput** implementations:
* **AnsiConsoleOutput** - implements IConsoleOutput methods using Ansi Escape codes
* **DefaultNetConsoleOutput** - implements IConsoleOutput methods using dotnet Console.XXXX Methods.
* **WindowsLegacyConsoleOutput** - extends DefaultNetConsoleOutput to use win32 api alternate screen buffer and underline which you can only do via pInvoke win32 apis.

All of IConsole implementations derive from ConsoleBase which wraps an IConsoleOutput allowing us to gracefully switch output between Ansi/API without having to change the input mechanisms. 

For example for windows support we do either 
```new Win32Console(new AnsiConsoleOutput())```
or
```new Win32Console(new WindowsLegacyConsoleOutput())```

Similarly for cross plat we could do:
```new DefaultNetConsoleOutput(new AnsiConsoleOutput())```
or
```new DefaultNetConsoleOutput(new DefaultNetConsoleOutput())```
etc. 

By default AnsiConsole() uses AnsiConsoleOutput and DefaultNetConsole uses DefaultNetConsoleOutput.

Changed AutoConsole() method to detect if we are running inside WindowsTerminal or cmd.exe, using AnsiConsoleOutput or DefaultNetConsoleOutput as appropriate. 

>NOTE: The upper left terminal is AnsiConsoleOutput running on Win32 Windows Terminal.
> The lower right terminal is LegacyWindowsConsoleOutput using purely Win32 APIs (no ANSI codes at all) using Ega ConsoleColor and running on cmd.exe. 
> It is auto detecting the appropriate console and color mode.
legacy

### CODE REVIEW

File: src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs [hunk 1]
```diff
@@ -0,0 +1,170 @@
...
+    ///     Console implementation which uses ANSI escape sequences for output
+    /// </summary>
+    public class AnsiConsoleOutput : IConsoleOutput
+    {
+        private const string TestEmoji = "👨‍👩‍👧‍👦";
+        private PixelBufferCoordinate _headBufferPoint;
+
+        private bool? _supportEmoji;
+
+        public bool CaretVisible { get; private set; }
```
jinek, line 22: I don't see CaretVisible property is used anywhere

src/Consolonia.Core/Infrastructure/DefaultNetConsole.cs
```diff
@@ -41,29 +46,20 @@ public class DefaultNetConsole : InputLessDefaultNetConsole
             (ConsoleModifiers.Shift, RawInputModifiers.Shift), (ConsoleModifiers.Alt, RawInputModifiers.Alt)
         ]);
 
-        public DefaultNetConsole()
-        {
-            StartSizeCheckTimerAsync();
-            StartInputReading();
-        }
+        public override bool SupportsAltSolo => true;
```
jinek, line 49: Console does not recognise ALT press, correct me if I'm wrong

src/Consolonia.GuiCS/WindowsDriver.cs
```diff
@@ -2,294 +2,63 @@
...
-			public byte R;
-			[FieldOffset (1)]
-			public byte G;
-			[FieldOffset (2)]
-			public byte B;
-
-			[FieldOffset (0)]
-			public uint Value;
-		}
+    [SupportedOSPlatform("windows")]
```
jinek, line 11: Just curios what does it do?

src/Consolonia.GuiCS/WindowsDriver.cs
```diff
@@ -2,294 +2,63 @@
...
+        {
+            const int bufferSize = 1;
+            var records = new INPUT_RECORD[bufferSize];
+            try
+            {
+                Kernel32.ReadConsoleInput(InputHandle, records, bufferSize,
+                    out var numberEventsRead);
+
+                return numberEventsRead == 0
+                    ? null
```
jinek, line 55: Better to return empty array?

src/Consolonia.GuiCS/WindowsDriver.cs
```diff
@@ -2,294 +2,63 @@
...
+                    : records;
+            }
+            catch (Exception)
```
jinek, line 58: Exception should be specified. If it's Win32 or COM better to check code too

src/Consolonia.GuiCS/WindowsDriver.cs
tomlm: If the class is used in code which can be accessed by cross platform it will give you a warning/error at build time. You then can disable the warning if you know it's ok. For example:
```csharp
if (Environment.Windows)
    console = new Win32Console(...);
```

src/Consolonia.GuiCS/WindowsDriver.cs
tomlm: hm. this was the original Gui.cs code, switching to  throw GetLastError().GetException();

src/Consolonia.Core/Infrastructure/DefaultNetConsole.cs
tomlm: you are correct.

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
tomlm: yeah, huh. I have removed.

src/Consolonia.GuiCS/WindowsDriver.cs
tomlm: yep

jinek (approved):

PR #245, opened 2025-01-04
Fix dialog title

tomlm:
* MessageBox shadow was broken
* Fixed title that overlapped close button
* Added a bit of turbovision pizazz to double line with ```==| text |===```
image


jinek (approved):

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -159,8 +159,7 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
                     // if we have strokes to draw
                     if (streamGeometry.Strokes.Count > 0)
                     {
-                        if (pen == null || pen.Thickness == 0)
-                            return;
+                        pen = pen ?? new Pen(brush);
```
jinek, line 162: What is the case when pen is not passed?

PR #250, opened 2025-01-07
Does not try to parse enum if it's not defined

jinek:


tomlm (commented):
I don't understand what you are trying to do here.

jinek:
> I don't understand what you are trying to do here.
> 

Same. I will close this pr

PR #251, opened 2025-01-12
Replace ConsoloniaApplication with 100% vanilla Avalonia Application

tomlm:
App should derive from ConsoloniaApplication
Bumped the version.


tomlm:
Do you have any ideas on how we can have the project reference tied to the build in the template?

src/Consolonia.Gallery/App.cs
```diff
@@ -1,12 +1,14 @@
...
+using Avalonia;
+using Avalonia.Controls.ApplicationLifetimes;
 using Consolonia.Core.Infrastructure;
 using Consolonia.Gallery.View;
 using Consolonia.Themes;
 
 namespace Consolonia.Gallery
 {
-    internal class App : ConsoloniaApplication<ControlsListView>
+    internal class App : Application
```
jinek, line 11: Why don't we reuse ConsoloniaApplication?

src/Consolonia.Gallery/App.cs
```diff
@@ -25,7 +27,12 @@ public override void OnFrameworkInitializationCompleted()
             else
                 Styles.Add(new TurboVisionTheme());
 
+            if (ApplicationLifetime is IClassicDesktopStyleApplicationLifetime desktop)
+                desktop.MainWindow = new ControlsListView();
+
             base.OnFrameworkInitializationCompleted();
+
+            this.InitializeConsolonia();
```
jinek, line 35: It's supposed to be automatic with no user intervention, right?

src/Tools/Consolonia.PreviewHost/App.cs
```diff
@@ -73,6 +74,8 @@ public override void OnFrameworkInitializationCompleted()
                     };
 
                 base.OnFrameworkInitializationCompleted();
+
+                this.InitializeConsolonia();
```
jinek, line 78: We could call it from base class, but now spreading to all inheritors, why is that?

src/Tools/Consolonia.Templates/ConsoloniaAppTemplate/ConsoloniaAppTemplate.csproj
```diff
@@ -8,7 +8,7 @@
     </PropertyGroup>
     
     <ItemGroup>
-        <PackageReference Include="Consolonia" Version="11.2.1-alpha-606" />
+        <PackageReference Include="Consolonia" Version="11.2.1-alpha-748" />
```
jinek, line 11: I think we need to research what we could do here eventually

jinek (approved):

Issue #253, opened 2025-01-13
App blows up if you set Windows.WindowStartupLocation


tomlm:
Throwing excepton saying that rendering system is not initialized.

Expected:
 The value is ignored.

PR #258, opened 2025-01-15
Fix WindowsStartupLocation throwing exception

tomlm:
Bug: #253 

When specificing WindowStartupLocation an exception is thrown saying "Rendering system is not intialized"

Avalonia expects there to be a IScreenImpl available so it can calculate things like window position. This delta implements a single
virtual screen for Avalonia to use to calculate.

It ignores request to resize and move the console window, although we could do something like that using platform APIs on some platforms (like windows).


src/Consolonia.Gallery/View/ControlsListView.axaml
```diff
@@ -4,6 +4,7 @@
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         mc:Ignorable="d"
+        WindowStartupLocation="CenterScreen"
```
jinek, line 7: So what does it do eventually, what is the effect?

tomlm: it doesn't blow up. 

As I told you in other PR, I've been experimenting with adding .Console platform projects to existing avalonia projects.  (A UI is shared between Desktop, Console, Android, etc)

WindowStartupLocation is a hint to the platform. On Desktop it positions the window. It's ignored by Android, and ignored by Console app.

jinek: Ah, understood, then add //todo: raise ConsoloniaNotSupportedException
please.

tomlm: Where would I leave it? It's Avalonia that throws because there the environment isn't set up correctly.  It expects there to be Screens defined by the platform so it can do screen based calculations. 
This PR implements the missing platform Screens functionality. 

The other place is that it generates a request to Resize() and Move() on the window. 
The current code for Resize looks like this:
```csharp 
  public void Resize(Size clientSize, WindowResizeReason reason = WindowResizeReason.Application)
  {
      // todo: can we deny resizing? TODO We could consider resizing the console window
  }
```
and I changed Move to be same
```csharp
  public void Move(PixelPoint point)
  {
      // ignore move request, TODO We could consider moving the console window
  }
```

jinek (approved):

PR #275, opened 2025-01-23
Doubleclick doesn't work anymore 

tomlm:
When we consolidated the code for dispatching inputs we started sending event timestamps as ticks, not milliseconds.  The avalonia code is expecting timestamp to be units of milliseconds and calculates double click period by milliseconds between clicks. The delta of ticks based timestamps was way larger than the millisecond value for double click detection and so no translation to double click was happening.

This fix changes code from using Stopwatch.GetTimestamp() to DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()


jinek:
We should not use `Now` because it is system dependent. For instance if clock synchronization happened during key press the key can stay ON "forever"

tomlm:
I switched to **Environment.TickCount64**
https://learn.microsoft.com/en-us/dotnet/api/system.environment.tickcount64?view=net-9.0

Definition: 
> Gets the number of milliseconds elapsed since the system started.

jinek (approved):

PR #280, opened 2025-01-25
Drag and drop works now

tomlm:
* Win32 was not modeling click drag the way Avalonia expects. Win32 drive generates MOVE/DOUBLECLICK, which we were mapping to MOVE and UP events.  Now we map to the way that Avalonia expects, which is a MOUSE BUTTON DOWN EVENT followed by MOVE with MOUSE_BUTTON_DOWN
 
* Added gallery page for drag drop
dragAndDrop


tomlm:
NOTE: I accidently did this on top of the branch with the ScrollBar fixes. Do not merge this before the other PR

PR #282, opened 2025-01-25
Drag and drop works now

tomlm:
* Win32 was not modeling click drag the way Avalonia expects. Win32 drive generates MOVE/DOUBLECLICK, which we were mapping to MOVE and UP events.  Now we map to the way that Avalonia expects, which is a MOUSE BUTTON DOWN EVENT followed by MOVE with MOUSE_BUTTON_DOWN
 
* Added gallery page for drag drop
dragAndDrop


tomlm:
NOTE: Linux is not working correctly because the curses driver is not generating mouse move events, only position information when the mouse button is pressed.  I've not been able to figure out why that is. Maybe something you can think of?

jinek:
> Maybe something you can think of?

We assume that linux terminals do not support mouse move event: `IConsole.SupportsMouseMove`

jinek (approved):

Issue #281, opened 2025-01-25
Feature: character based Cursor pointer


tomlm:
When we are doing operations like Drag and Drop Avalonia normally changes the mouse pointer to reflect information like Copy/Move/Delete/Invalid to drop etc. 

We don't have a cursor.  We should consider if we want to have a character based pointer for the mouse so we can support avalonia pointer concepts.

```csharp
public enum StandardCursorType
    {
        Arrow,
        Ibeam,
        Wait,
        Cross,
        UpArrow,
        SizeWestEast,
        SizeNorthSouth,
        SizeAll,
        No,
        Hand,
        AppStarting,
        Help,
        TopSide,
        BottomSide,
        LeftSide,
        RightSide,
        TopLeftCorner,
        TopRightCorner,
        BottomLeftCorner,
        BottomRightCorner,
        DragMove,
        DragCopy,
        DragLink,
        None,
        
        // Not available in GTK directly, see https://www.pixelbeat.org/programming/x_cursors/
        // We might enable them later, preferably, by loading pixmax directly from theme with fallback image
        // SizeNorthWestSouthEast,
        // SizeNorthEastSouthWest,
    }
```

PR #309, opened 2025-03-24
Implement console based software cursor.

tomlm:
consoleCursor

consoleCursor2

implements #281


tomlm:
Well, I dunno, are there terminals with mouse input but no cursor?

I went back and forth on that....

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -469,5 +482,118 @@ protected virtual void Dispose(bool disposing)
...
             }
         }
+
+        /// <summary>
+        ///     This works by creating a Software cursor, aka a sprite that is drawn on top of the screen.
+        ///     It draws the current cursor directly to the console buffer, but maintains a list of the pixels need to be redrawn
+        ///     whenever the cursor moves.
+        /// </summary>
+        /// <param name="point"></param>
+        private void RenderSoftwareCursor(Point point)
```
jinek, line 492: Would not be simplier to draw a symbol in the overlay?

tomlm: I tried that first, all sorts of issues with clipping, getting in the way of eventing, etc

jinek:
> Well, I dunno, are there terminals with mouse input but no cursor?
> 
> I went back and forth on that....

Can we remove default cursor or is there a reason to keep? I don't see a point to have cursor under mouse system cursor.

src/Consolonia.Core/Infrastructure/ConsoleCursorFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+using Avalonia.Platform;
+
+namespace Consolonia.Core.Infrastructure
+{
+#pragma warning disable CA1711 // Identifiers should not have incorrect suffix
+    internal sealed class ConsoleCursorFactory : ICursorFactory
+    {
+        private readonly Dictionary<StandardCursorType, CursorImpl> _standardCursors = new();
+
+        public ICursorImpl GetCursor(StandardCursorType cursorType)
```
jinek, line 14: Just in case you don't know. you can simply mark method with [MethodImpl(Synchronized)]

src/Consolonia.Core/Infrastructure/ConsoleCursorFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+                    return cursor;
+                cursor = new CursorImpl(cursorType);
+                _standardCursors.Add(cursorType, cursor);
+                return cursor;
+            }
+        }
+
+        public ICursorImpl CreateCursor(IBitmapImpl cursor, PixelPoint hotSpot)
+        {
+            throw new NotSupportedException("Consolonia doesn't support bitmap based cursors");
```
jinek, line 28: I think it's a good place to RaiseNotSupported. But again, we need to improve it for possibility to return things, not just throwing exceptions.It's just for future discussions

src/Consolonia.Core/Infrastructure/ConsoleCursorFactory.cs
tomlm: What does that do?

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -384,6 +394,9 @@ private void ConsoleOnMouseEvent(RawPointerEventType type, Point point, Vector?
                         (Vector)wheelDelta!, modifiers));
                     break;
             }
+
+            // draw the software cursor at this mouse position
+            RenderSoftwareCursor(point);
```
jinek, line 399: Just out of curiosity: why are we thinking, that no render will happen after mousemove event? Becuase it's "input" priority in the dispatcher?

jinek (approved):

src/Consolonia.Core/Infrastructure/ConsoleCursorFactory.cs
jinek: it locks by the current instance which contains this method

Issue #304, opened 2025-03-23
Check 7bit coloring


jinek:
Some terminals support only 7bit - for them bright variant of background is not available.
TurboVision theme must be repainted so it does not rely on bright background.
As an example check Menu:

![Image](https://github.com/user-attachments/assets/31a2df1f-25f1-425f-be79-0b838f44dd23)

jinek:
Additionally need to check TreeView
![Image](https://github.com/user-attachments/assets/07570f27-6486-481f-b6c9-5d631f986044)

And drop down (on some specific page)
![Image](https://github.com/user-attachments/assets/2232367d-4880-4227-a9d1-e64f5aebbb99)

jinek:
As a reference

![Image](https://github.com/user-attachments/assets/50772682-e118-4373-8303-56b8f21a0e1c)

jinek:
![Image](https://github.com/user-attachments/assets/e41c1aff-a7a9-4e17-a3a9-48ed3b22d21a)

jinek:
![Image](https://github.com/user-attachments/assets/ba943aa8-cc6e-43f5-a876-fb63e793aa80)

![Image](https://github.com/user-attachments/assets/ad2de56e-94f5-4d93-b153-f0d28fbabfc7)

![Image](https://github.com/user-attachments/assets/c2a4e02b-87d7-43c0-87dd-491be76f6af9)

jinek:
Others issues will be created separately

PR #314, opened 2025-03-29
304 check 7bit coloring

jinek:
#304


tomlm:
Calendar looks broken, button layout of right button
image

image

tomlm:
Cursor for listbox flashs over the top of the scroll bar
image
1. start app
2. hit tab key

tomlm:
Button spinner on fluent looks weird.
image
Can't we use better up/down symbols?

tomlm:
window icons are weird on turbo vision windows
image

tomlm:
Window icons are weird on fluent windows
image

tomlm:
image
Expander is using default button style which has borders which causes 
* Expander to be 3 chars high
* Borders are being rendered inside the content area
I think we should disable borders in this case.

tomlm:
Radio button uses Block caret, Combobox uses underscore on fluent theme
image
Seems like they should be the same.

tomlm:
Slider should position caret with the selector, not the range
image
image

tomlm:
disabled button on fluent is unreadable.
image

tomlm:
horizontal scrollviewer has white start and end char, but vertical doesn't

image

tomlm:
transitioning control is built using buttons, which means it has large edge boundaries.

It looks weird to me. 

image

tomlm:
weird curses bug? I can't repo this on windows:
weird

1. start gallery
2. scroll using wheel mouse, at some point it renders the list with gaps in it
3. keep scrolling and the gaps get filled, 
4. scroll back and the gaps are back.
It seems like it might be sensitive to the size of the window, try sizing it so that "MENU" is the last item before you start scrolling.

tomlm:
One last comment. I think we should consider that Fluent is OK to use full spectrum of Unicode chars and then keep TurboVision as theme to use for simpler consoles without unicode support.

Does that make sense?

jinek:
> One last comment. I think we should consider that Fluent is OK to use full spectrum of Unicode chars and then keep TurboVision as theme to use for simpler consoles without unicode support.
> 
> Does that make sense?

Yes, I thought we already agreed on this. Why are you mentioning now? Is it smth I missed?

src/Consolonia.Themes/Fluent/FluentTheme.axaml
```diff
@@ -8,15 +8,15 @@
                 <ResourceInclude Source="/Fluent/FluentColors.axaml" />
             </ResourceDictionary.MergedDictionaries>
 
-            <x:String x:Key="ThemeLeftArrowSymbol">˂</x:String>
-            <x:String x:Key="ThemeRightArrowSymbol">˃</x:String>
-            <x:String x:Key="ThemeUpArrowSymbol">˄</x:String>
-            <x:String x:Key="ThemeDownArrowSymbol">˅</x:String>
+            <x:String x:Key="ThemeLeftArrowSymbol">&lt;</x:String>
```
tomlm, line 11: maybe keep the unicode chars for fluent and use these for Turbo?

src/Consolonia.Themes/Fluent/FluentTheme.axaml
```diff
@@ -42,12 +42,19 @@
             <x:String x:Key="ThemeCheckedSymbol">✓</x:String>
             <x:String x:Key="ThemeCloseSymbol">✕</x:String>
 
-            <x:String x:Key="ThemeToggleSwitchOffSymbol">◯</x:String>
-            <x:String x:Key="ThemeToggleSwitchOnSymbol">⬤</x:String>
+            <x:String x:Key="ThemeToggleSwitchOffSymbol">●</x:String>
```
tomlm, line 45: I hate losing these theme characters. same comment as above?

tomlm (approved):
I would love to not use the sexier unicode chars. Do you think we need a different theme on top of Fluent one which uses unicode more extensively for things like arrows?

src/Consolonia.Themes/Fluent/FluentTheme.axaml
jinek: They just looked but, I thought this is what we discussed here:
image
Currently in fluent theme it looks like this:
image
image
image
Or you think it's better?

src/Consolonia.Themes/Fluent/FluentTheme.axaml
jinek: I just tried to match Avalonia:
image
it's filled black when off and filled white when on.
Should I revert?

jinek:
> disabled button on fluent is unreadable. image

Fixed

jinek:
> weird curses bug? I can't repo this on windows: weird [ weird ](https://private-user-images.githubusercontent.com/17789481/429567537-5b305440-564e-48b2-b253-c47985f65a6a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDQ0NDE4ODQsIm5iZiI6MTc0NDQ0MTU4NCwicGF0aCI6Ii8xNzc4OTQ4MS80Mjk1Njc1MzctNWIzMDU0NDAtNTY0ZS00OGIyLWIyNTMtYzQ3OTg1ZjY1YTZhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA0MTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNDEyVDA3MDYyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTIzZGZkNWFiM2NiMzM0ZDg1MTAyNjk5MTE3ZjIyZDE5NGU3NzkwZjk0ZDdkMTE5N2M3ODU3OWZhYmEyYjM0ZTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.fnR2NjbZY6_DCG72EvHpzflWz1q475FaOaoLYTRgslw) [ ](https://private-user-images.githubusercontent.com/17789481/429567537-5b305440-564e-48b2-b253-c47985f65a6a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDQ0NDE4ODQsIm5iZiI6MTc0NDQ0MTU4NCwicGF0aCI6Ii8xNzc4OTQ4MS80Mjk1Njc1MzctNWIzMDU0NDAtNTY0ZS00OGIyLWIyNTMtYzQ3OTg1ZjY1YTZhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA0MTIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNDEyVDA3MDYyNFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTIzZGZkNWFiM2NiMzM0ZDg1MTAyNjk5MTE3ZjIyZDE5NGU3NzkwZjk0ZDdkMTE5N2M3ODU3OWZhYmEyYjM0ZTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.fnR2NjbZY6_DCG72EvHpzflWz1q475FaOaoLYTRgslw)
> 
>     1. start gallery
> 
>     2. scroll using wheel mouse, at some point it renders the list with gaps in it
> 
>     3. keep scrolling and the gaps get filled,
> 
>     4. scroll back and the gaps are back.
>        It seems like it might be sensitive to the size of the window, try sizing it so that "MENU" is the last item before you start scrolling.
> 
> 
> I can only repo using scroll wheel. PageUp down keys and clicking on the scroll bar don't seem to have issue.

This one is reproducable in main branch: 
image

src/Consolonia.Themes/Fluent/FluentTheme.axaml
jinek: I've moved this to discuss in a separate PR https://github.com/jinek/Consolonia/pull/324

src/Consolonia.Themes/Fluent/FluentTheme.axaml
jinek: Also moved to separate PR: https://github.com/jinek/Consolonia/pull/324

jinek:
> Cursor for listbox flashs over the top of the scroll bar

fixed

jinek:
> Cursor gets left on after redraw?

fixed

jinek:
> window icons are weird on turbo vision windows image

Will address this in a separate PR. Turbovision look should be based on something like this:
image

jinek:
> image Expander is using default button style which has borders which causes
> 
>     * Expander to be 3 chars high
> 
>     * Borders are being rendered inside the content area
>       I think we should disable borders in this case.

Fixed

jinek:
> Radio button uses Block caret, Combobox uses underscore on fluent theme image Seems like they should be the same. I think I like the block cursor

Now it's same on ComboBox and RadioButton.

jinek:
> Slider should position caret with the selector, not the range image image
> 
> This would cause the selector to block blink while it has the focus, which seems better to me

This will address in next chunk

jinek:
> transitioning control is built using buttons, which means it has large edge boundaries.
> 
> It looks weird to me.
> 
> image
> 
> Should we remove boundary?

fixed

jinek:
> Fluent treeview colors are different when opened image
> 
> Turbo color stays same image
> 
> Also arrows on are different styles, triangle vs arrow.

TreeView will fully address in next chunk

tomlm:
Looks like canvas lines are going 1 square too far?
image

tomlm:
This is not a bug, but I wonder if grid splitter should have a single line
image

tomlm:
Spacing on messagebox seems big
image

tomlm:
Colors for turbovision on expander seem wonkyk to me
image

tomlm:
Cursor for showing focus seems weird using invetered blinking. does _ work?
image

tomlm (dismissed):

jinek:
> Hmmm...not crazy with how the header on this looks, buttons overlap and 3 char high? image

This answered here https://github.com/jinek/Consolonia/pull/314#issuecomment-3037529649
Will fix windows in separate PR.

jinek:
> turbovision messagebox feels "heavy" to me image

Yeah, button is now having internal padding left and right 1px, comparing to how it was:
image
But with padding OK buttons looks better, no?

jinek:
> Carousel with TurboVision looks weird with drop shadow image

It was same, I'm touching it in this PR.

jinek:
> Cursor for showing focus seems weird using invetered blinking. does _ work? image

This answere here https://github.com/jinek/Consolonia/pull/314#issuecomment-3039810304
This control will address in next PR.

PR #324, opened 2025-06-08
Fluent symbols

jinek:


jinek:
Addressed in https://github.com/jinek/Consolonia/pull/390

PR #332, opened 2025-07-27
Fix Alt+Tab focus bug

jinek:
## Summary
- fix alt key sequence parsing to use the actual key

## Testing
- `dotnet test src/Consolonia.sln -v minimal`

------
https://chatgpt.com/codex/tasks/task_e_6886989df83883228c2774e98473af61


jinek:
Does not work

PR #395, opened 2025-08-17
Switch back to IClassicApplicationLifetime

tomlm:
We have come full circle, we were using 
* **ISingleViewApplicationLifetime** which added **MainView** property
* **ISingleTopLevelApplication** which added **TopLevel** property
But we evolved to require **TopLevel** to be a **Window** and **MainView**  to the contents of the TopLevel window.

That's **IClassicDesktopApplicationLifeTime**. 

This PR simply switches back to **IClassicDesktopApplicationLifetime**
* **IClassicDesktopApplicationLifeTime** defines **MainWindow** which is a Window at the root.
* We remove **TopLevel** and **MainView**

Cleanup all references, update template definition etc.

Unit tests Pass, Gallery Works, Example works, BlazorExample works.


src/Consolonia.Core/ConsoloniaLifetime.cs
```diff
@@ -47,26 +44,11 @@ public void Shutdown(int exitCode = 0)
...
-                {
-                    if (TopLevel == null)
-                        TopLevel = new Window();
-                    TopLevel.Content = value;
-                }
-            }
-        }
+        public Window MainWindow { get; set; }
+
+        public IReadOnlyList<Window> Windows => new List<Window> { MainWindow };
```
jinek, line 51: If we are accessing or supposed to access this list frequently, I would propose to instantiate just an array at the moment of MainWindow.set .But otherwise does not matter

src/Example/Views/DataGridView.axaml
```diff
@@ -1,108 +1,107 @@
...
-                        <StackPanel.Styles>
-                            <Style Selector="TextBlock">
-                                <Setter Property="Foreground"
-                                        Value="Gray" />
-                            </Style>
-                        </StackPanel.Styles>
-                        <TextBlock Text="Id" />
-                        <TextBlock Text="Title" />
-                        <TextBlock Text="Genre(ALT+DOWN)" />
+<Grid xmlns="https://github.com/avaloniaui"
```
jinek, line 1: If you are making any other changes to this PR. Lets drop this project, it's nonsense now.

PR #399, opened 2025-08-19
Rename Fluent theme to ModernTheme

tomlm:
Rename all of the things.

> NOTE: There is still a FluentTheme defined so we don't break any code that exists that expects the old theme name. It has a comment to delete in a near future release.


src/Consolonia.Themes/ModernTheme.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class ModernTheme : Styles
+    {
+        public ModernTheme(IServiceProvider sp = null)
+        {
+            AvaloniaXamlLoader.Load(sp, this);
+        }
+    }
+
+    // TODO: Remove this class when FluentTheme is no longer needed.
```
jinek: [Obsolete()] ?

jinek (dismissed):

src/Consolonia.Themes/ModernTheme.cs
tomlm: good idea

jinek (approved):

PR #400, opened 2025-08-19
Upgrade Avalonia version to 11.3.4

tomlm:
Update package versions, bump version.

2 issues
* IDrawingContextImpl needs to implement PushRenderOptions()/PopRenderOptions() (at least not throw not implemented)
* IClipboard interface has some extra methods on it.
* IStorage interface has some extra methods to implement

### DIFF
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -57,6 +57,7 @@ namespace Consolonia.Core.Drawing
         private readonly PixelBuffer _pixelBuffer;
         private readonly Matrix _postTransform = Matrix.Identity;
         private Matrix _transform = Matrix.Identity;
+        private readonly Stack<RenderOptions> _renderOptions = new Stack<RenderOptions>();
 
         public DrawingContextImpl(ConsoleWindowImpl consoleWindowImpl)
         {
@@ -351,12 +352,12 @@ namespace Consolonia.Core.Drawing
 
         public void PushRenderOptions(RenderOptions renderOptions)
         {
-            throw new NotImplementedException();
+            _renderOptions.Push(renderOptions);
         }
 
         public void PopRenderOptions()
         {
-            throw new NotImplementedException();
+            _renderOptions.Pop();
         }
 
         public object GetFeature(Type t)
--- a/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
+++ b/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
@@ -502,5 +502,15 @@ namespace Consolonia.Core.Infrastructure
         {
             CursorChanged?.Invoke(obj);
         }
+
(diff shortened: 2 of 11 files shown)


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -56,6 +56,9 @@ internal class DrawingContextImpl : IDrawingContextImpl
         private readonly ConsoleWindowImpl _consoleWindowImpl;
         private readonly PixelBuffer _pixelBuffer;
         private readonly Matrix _postTransform = Matrix.Identity;
+
+        // ReSharper disable once CollectionNeverQueried.Local
+        private readonly Stack<RenderOptions> _renderOptions = new();
```
jinek, line 61: But we are not using them anywhere, why to stack?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -351,12 +354,12 @@ public void PopGeometryClip()
 
         public void PushRenderOptions(RenderOptions renderOptions)
         {
-            throw new NotImplementedException();
+            _renderOptions.Push(renderOptions);
```
jinek, line 357: Why did it start to throw?

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -331,6 +331,16 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
+        public void SetCanMinimize(bool value)
+        {
+            // ignore, we can't
```
jinek, line 336: Why is it requested? Should it be `RaiseNotSupported` ? Same for SetCanMaximize.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: Well, if we need it we've been collecting it  I'm fine not collecting it too

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: It would get called when running unit tests. I don't know, old version of avalonia didn't call it, new one does.

src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
tomlm: app won't run, it always calls this.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: I have remved the stack

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: I have removed the stack

PR #419, opened 2025-08-29
fix simple rect calc and add a Label style

tomlm:
If Label is not in a container which stretches then it has a style with a padding of 6.

### DIFF
--- a/src/Consolonia.Themes/Templates/Controls/Label.axaml
+++ b/src/Consolonia.Themes/Templates/Controls/Label.axaml
@@ -1,7 +1,23 @@
 <ResourceDictionary xmlns="https://github.com/avaloniaui"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                     x:ClassModifier="internal">
-    <Style Selector="Label">
-        <Setter Property="Padding" Value="0 0 1 0"/>
-    </Style>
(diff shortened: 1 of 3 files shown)


src/Consolonia.Themes/Templates/Controls/Label.axaml
```diff
@@ -0,0 +1,24 @@
+<ResourceDictionary xmlns="https://github.com/avaloniaui"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    x:ClassModifier="internal">
+    <ControlTheme x:Key="{x:Type Label}"
+                  TargetType="Label">
+        <Setter Property="Padding"
+                Value="0 0 1 0" />
```
jinek, line 7: Why do we need default padding from right?

tomlm: The original template has 6 pixel right border to create space between the label and the input control.

tomlm: https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Themes.Simple/Controls/Label.xaml

tomlm: Padding of three

PR #424, opened 2025-09-04
RaiseNotImplemented for ConsoleWindowImpl

tomlm:
Ran into this while porting code.

### DIFF
--- a/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
+++ b/src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
@@ -153,18 +153,17 @@ namespace Consolonia.Core.Infrastructure
 
         public void Hide()
         {
-            // console window can't do this.
+            ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.ConsoleWindowHideNotSupported);
         }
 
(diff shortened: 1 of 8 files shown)


src/Consolonia.Core/Infrastructure/ConsoleWindow.cs
```diff
@@ -153,14 +153,12 @@ public void Show(bool activate, bool isDialog)
 
         public void Hide()
         {
-            // toplevel never hides
-            throw new NotImplementedException();
+            // console window can't do this.
```
jinek: Lets better call `ConsoloniaPlatform.RaiseNotSupported` and add ignore at https://github.com/jinek/Consolonia/blob/c13a8b28846b9da01823b66087d0e84644397bd6/src/Consolonia.Core/Infrastructure/ConsoloniaPlatform.cs#L130
We can have one single enum value for all these cases.

jinek: This way it will be ignored by default, but if needed will be possible to investigate why it does not work

jinek (approved):

jinek (approved):

PR #427, opened 2025-09-07
Tabs were messed up

tomlm:
* MeasureText returned for \t -1 but we were casting to ushort, making it's width maxvalue
* We were rendering as spaces instead of just advancing the pixel buffer.

We are now aligned 100% with Avalonia behavior, TextBox with AcceptsTab and AvalaloniaEdit both work exactly as we desire.

### DIFF
--- a/src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
+++ b/src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
@@ -166,6 +166,7 @@ namespace Consolonia.Core.Drawing.PixelBufferImplementation
             return new Pixel(newForeground, newBackground, newCaretStyle);
         }
 
+        // ReSharper disable once InconsistentNaming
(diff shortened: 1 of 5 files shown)


src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs
```diff
@@ -166,6 +166,10 @@
             return new Pixel(newForeground, newBackground, newCaretStyle);
         }
 
+        // ReSharper disable once InconsistentNaming
```
jinek, line 169: Why do you want to change the naming style here?

tomlm: I prefer _ I guess

jinek: Ok, lets leave this 😅

jinek (approved):

PR #429, opened 2025-09-08
Add Consolonia.AvaloniaEdit project and clean up packages

tomlm:
* Add Accelerate dev tooling to gallery/consolonia edit
* Add Consolonia.AvaloniaEdit with theme file, making it 1 line to enable AvaloniaEdit
* Added Caret brush attached property to enable cursor based AvaloniaEdit caret.
* Cleaned up packages
  * Added unique description to each project file
  * Added assets/images folder as those links are "acceptable" by nuget for embedded images in the readme's
  * Cleaned up readmes to use appropriate imagery

### DIFF
--- a/src/Consolonia.AvaloniaEdit/Theme.axaml
+++ b/src/Consolonia.AvaloniaEdit/Theme.axaml
@@ -45,7 +45,7 @@
             <!-- There is a bug in AvaloniaEdit which hard codes line height to 1.35 * FontSize. To get line height of 1 we set FontSize to 0.7407407407407407 of desired size-->
             <Setter Property="FontSize"
                     Value="0.7407407407407407" />
-            
+
             <!-- If AvaloniaEdit ever supports CaretBrush as an avalonia property we will do something like this:
             <Style Selector="^ /template/ editing|TextArea[OverstrikeMode=false]">
                 <Setter Property="CaretBrush">
@@ -72,7 +72,7 @@
                  Brush="Transparent" />
         </Setter>
     </Style>
-    
+
     <!--disable border around selections-->
     <Style Selector="editing|TextArea">
         <Setter Property="SelectionBorder">
@@ -80,6 +80,6 @@
                  Brush="Transparent" />
         </Setter>
         <Setter Property="SelectionCornerRadius"
-                Value="0"/>
+                Value="0" />
     </Style>
 </Styles>


assets/images/AvaloniaEdit.gif
jinek, line 1: We should not store large media files in the repo. I guess logos are fine, but gif size is comparable to entire repo size. Download/upload speed would be 2x.

src/Consolonia.AvaloniaEdit/Caret.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    /// <summary>
+    ///     This attached property is used to enable console style caret in AvaloniaEdit TextEditor.
+    /// </summary>
+    /// <remarks>
+    ///     It is autoamtically added using the Consolonia.AvaloniaEdit Theme setter
+    ///     It can be explitily added to a text editor XAML using: console:Caret.UseConsole="True"
+    ///     It can be explicitly added to a text editor in code using: Caret.SetUseConsole(textEditor, true);
+    /// </remarks>
+    public sealed class Caret
```
jinek, line 17: Would not be easier to setup both properties via a style? I guess it would be twice shorter even together with closing tags.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -239,10 +240,16 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
                             () =>
                             {
-                                _pixelBuffer.Set((PixelBufferCoordinate)head,
-                                    pixel => pixel.Blend(new Pixel(moveBrush.CaretStyle)));
+                                Pixel pixel = _pixelBuffer[(PixelBufferCoordinate)head];
+                                if (pixel.CaretStyle != moveBrush.CaretStyle)
+                                {
+                                    // only be dirty if something changed
+                                    _consoleWindowImpl.DirtyRegions.AddRect(
+                                        CurrentClip.Intersect(new Rect(head, new Size(1, 1))));
```
jinek: Here are no need to intersect with CurrentClip because this entire delegate would not be executed if head is out of it.

src/Consolonia.Gallery/Program.cs
```diff
@@ -19,6 +19,7 @@ public static AppBuilder BuildAvaloniaApp()
                 .UseConsolonia()
                 .UseAutoDetectedConsole()
                 .ThrowOnErrors()
+                .WithDeveloperTools()
```
jinek, line 22: But they still don't work if we don't remove ThrowOnErrors above, right?

src/Experimental/Consolonia.Editor/MainWindow.axaml.cs
```diff
@@ -166,9 +161,12 @@
 
         private void Caret_PositionChanged(object sender, EventArgs e)
         {
-            _statusTextBlock.Text = string.Format("Line {0} Column {1}",
-                _textEditor.TextArea.Caret.Line,
-                _textEditor.TextArea.Caret.Column);
+            if (_statusTextBlock != null)
```
jinek: Just curios why it becomes null?

src/Consolonia.Gallery/Program.cs
tomlm: no it works if we remove throwonerrors

src/Consolonia.AvaloniaEdit/Caret.cs
tomlm: Caret is not a dependency object, it's a POCO with no change notification or anything and the stylesheet can't access it. That's why I'm jumping through this hoop.

assets/images/AvaloniaEdit.gif
tomlm: I moved it to tomlm/ConsoloniaContent repo

src/Consolonia.AvaloniaEdit/Caret.cs
jinek: I mean you can set the caret style like this:

<Theme>
   <Style Selector="^[OverstrikeMode=false] /template/ TextArea">
      <Setter Property="CaretBrush" Value="{MoveConsoleCaretToPositionBrush SteadyBar}"/>
</Style>
</Theme>

Issue #140, opened 2024-11-08
On ubuntu when exiting Consolonia app terminal console no longer shows keystrokes


tomlm:
Repo steps
1. start gallery app
2. exit gallery app
3. try and use console.

tomlm:
This now shows keys, but mouse events are being pushed to the console.  I think the CursesConsole is not cleaning up registration for mouse events on exit.

tomlm:
#270 is correctly cleaning up console, but there is a dotnet bug which will be fixed in dotnet 10 where characters are not visible after exiting curses app.
https://github.com/dotnet/runtime/pull/111272

jinek:
> #270 is correctly cleaning up console, but there is a dotnet bug which will be fixed in dotnet 10 where characters are not visible after exiting curses app.
> https://github.com/dotnet/runtime/pull/111272 

Its now fixed, but I have reported another reoevant issue

https://github.com/dotnet/runtime/issues/128334

PR #431, opened 2025-09-11
fix linux console messed up on exit of consolonia app.

tomlm:
Fix for #140 
* Console.WriteLine( with escape codes makes them not work. use WriteText() not Console.WriteLine()
* Replaced Example with simple example so that it aligns with Example.Blazor


src/Consolonia.Core/Infrastructure/ConsoleBase.cs
```diff
@@ -27,6 +26,8 @@ protected ConsoleBase(IConsoleOutput consoleOutput)
             if (consoleOutput is ConsoleBase)
                 throw new ArgumentException("ConsoleBase cannot be used as a console output", nameof(consoleOutput));
 
+            Console.TreatControlCAsInput = true;
```
jinek, line 29: How did it work without this property?

PR #434, opened 2025-09-12
Reduce render time by ~30-50% by getting rid of lamdas and other perf tweaks

tomlm:
Using Avalonia Acceleartor metrics I was able to reduce render time by ~30-50% based on the situation by:
* Unwrapped ExecuteClip to just an if statement
* Got rid of PixeBuffer.Set() 

Also
* defined common Pixel, PixelForeground, PixelBackground, SimpleSymbol to avoid overhead of calling constructors for simple default operations.
* Added agressive inline hints to compiler for structs ctors
* got rid of stupid string alloc to represent bits in Image drawing <-- this was doing a bunch of memory allocs

I think lamda was most of perf benefit. I'm measuring ~30-50% faster render times, mostly when I have full screen refresh.  I basically made it so there was a continuous redraw of the entire screen via an animation and measured that.


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -532,24 +528,26 @@ private void FillRectangleWithBrush(IBrush brush, IPen pen, Rect r)
                 int py = (int)(r2.TopLeft.Y + y);
                 Color backgroundColor = brush.FromPosition(x, y, (int)width, (int)height);
 
-                CurrentClip.ExecuteWithClipping(new Point(px, py), () =>
+                if (CurrentClip.ContainsExclusive(new Point(px, py)))
```
jinek: With the delegate we can do var targetRectangle = new (r2.TopLeft, width, height); targetRectangle.Intersect(CurrentClip).Foreach(point=> ... );

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -494,20 +492,18 @@ private void ApplyTextDecorationLineInternal(IPen pen, Line line)
             for (int x = 0; x < line.Length; x++)
             {
                 Point h = head;
-                CurrentClip.ExecuteWithClipping(h, () =>
+                if (CurrentClip.ContainsExclusive(h))
```
jinek: We can run the loop thru included points only initially

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -720,15 +718,15 @@ private void DrawBoxPixelAndMoveHead(ref Point head, Line line, LineStyle lineSt
             for (int i = 0; i < count; i++)
             {
                 Point h = head;
-                CurrentClip.ExecuteWithClipping(h, () =>
+                if (CurrentClip.ContainsExclusive(h))
```
jinek: Same here: target line can be intersect at the beginning once

src/Experimental/Consolonia.Editor/MainWindow.axaml.cs
```diff
@@ -86,11 +86,15 @@
             if (!ApplyBrushAction(e, "editor.selectionBackground",
                     brush => _textEditor.TextArea.SelectionBrush = brush))
             {
-                if (Application.Current!.TryGetResource("TextAreaSelectionBrush", out var resourceObject))
+                if (!ApplyBrushAction(e, "editor.selectionHighlightBackground",
```
jinek, line 89: Not related?

tomlm: oh, yeah, that was a selection bug I found, I can pull out if you want

jinek: no need, just clarifying

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -126,12 +125,11 @@ public void DrawBitmap(IBitmapImpl source, double opacity, Rect sourceRect, Rect
                 var imagePixel = new Pixel(
                     new PixelForeground(new SimpleSymbol(quadPixel), foreground),
                     new PixelBackground(background));
-                CurrentClip.ExecuteWithClipping(new Point(px, py),
-                    () =>
-                    {
-                        _pixelBuffer.Set(new PixelBufferCoordinate((ushort)px, (ushort)py),
-                            existingPixel => existingPixel.Blend(imagePixel));
-                    });
+                if (CurrentClip.ContainsExclusive(new Point(px, py)))
```
jinek, line 128: Add todo here to later apply same trick to run only thru included points

jinek (approved):

PR #437, opened 2025-09-15
DO NOT MERGE- PixelBuffer using ref and Span<Pixel> with mutable structs

tomlm:
in theory this prevents copies and allocations. 

I didn't see any perf gains.


tomlm:
I don' t think we need to do this. With dropping of ISymbol we are getting the kind of struct value performance we want.

PR #440, opened 2025-09-18
Drop ISymbol so symbol structs are not on heap

tomlm:
* Get rid of ISymbol
* Merge **SimpleSymbol** and **DrawingBoxSymbol** into single struct **Symbol**
* Move Box character pattern logic into **BoxPattern** class.
* make all structs readonly fields
* Add **Character** ch and **Complex** string to Symbol so we have no string references 99% of the time.
* Update all references, unit tests, etc.

Inspecting the heap I see zero Symbol instances, 3000 strings and 1 pixel array of ~700k
image
image

### DIFF
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -162,33 +162,33 @@ namespace Consolonia.Core.Drawing
                     DrawLineInternal(pen, myLine);
                     break;
                 case StreamGeometryImpl streamGeometry:
+                {
+                    // if we have fills to do.
+                    if (streamGeometry.Fills.Count > 0)
+                        foreach (Rectangle fill in streamGeometry.Fills)
+                            DrawRectangle(brush, pen, new RoundedRect(fill.Rect));
+
+                    // if we have strokes to draw
+                    if (streamGeometry.Strokes.Count > 0)
                     {
-                        // if we have fills to do.
-                        if (streamGeometry.Fills.Count > 0)
-                            foreach (Rectangle fill in streamGeometry.Fills)
-                                DrawRectangle(brush, pen, new RoundedRect(fill.Rect));
+                        pen ??= new Pen(brush);
 
-                        // if we have strokes to draw
-                        if (streamGeometry.Strokes.Count > 0)
+                        RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
+                        for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
                         {
-                            pen ??= new Pen(brush);
-
-                            RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
-                            for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
-                            {
-                                Line stroke = streamGeometry.Strokes[iStroke];
-                                RectangleLinePosition strokePosition = strokePositions[iStroke];
-                                if (strokePosition == RectangleLinePosition.Left)
-                                    DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Left);
-                                else if (strokePosition == RectangleLinePosition.Right)
-                                    DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Right);
-                                else if (strokePosition == RectangleLinePosition.Top)
-                                    DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Top);
-                                else if (strokePosition == RectangleLinePosition.Bottom)
-                                    DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Bottom);
-                            }
+                            Line stroke = streamGeometry.Strokes[iStroke];
+                            RectangleLinePosition strokePosition = strokePositions[iStroke];
+                            if (strokePosition == RectangleLinePosition.Left)
+                                DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Left);
+                            else if (strokePosition == RectangleLinePosition.Right)
+                                DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Right);
+                            else if (strokePosition == RectangleLinePosition.Top)
+                                DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Top);
+                            else if (strokePosition == RectangleLinePosition.Bottom)
+                                DrawBoxLineInternal(pen, stroke, RectangleLinePosition.Bottom);
                         }
                     }
+                }
                     break;
                 default:
                     ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.DrawGeometryNotSupported, this, brush,
@@ -230,28 +230,28 @@ namespace Consolonia.Core.Drawing
                     case VisualBrush:
                         throw new NotImplementedException();
                     case ISceneBrush sceneBrush:
-                        {
-                            ISceneBrushContent sceneBrushContent = sceneBrush.CreateContent();
-                            sceneBrushContent?.Render(this, Matrix.Identity);
-                            return;
-                        }
+                    {
+                        ISceneBrushContent sceneBrushContent = sceneBrush.CreateContent();
+                        sceneBrushContent?.Render(this, Matrix.Identity);
+                        return;
+                    }
                     case MoveConsoleCaretToPositionBrush moveBrush:
+                    {
+                        Point head = r.TopLeft.Transform(Transform);
+                        if (CurrentClip.ContainsExclusive(head))
                         {
-                            Point head = r.TopLeft.Transform(Transform);
-                            if (CurrentClip.ContainsExclusive(head))
+                            Pixel pixel = _pixelBuffer[head];
+                            if (pixel.CaretStyle != moveBrush.CaretStyle)
                             {
-                                Pixel pixel = _pixelBuffer[head];
-                                if (pixel.CaretStyle != moveBrush.CaretStyle)
-                                {
-                                    // only be dirty if something changed
-                                    _consoleWindowImpl.DirtyRegions.AddRect(new Rect(head, new Size(1, 1)));
-                                    _pixelBuffer[head] =
-                                        pixel.Blend(new Pixel(moveBrush.CaretStyle));
-                                }
+                                // only be dirty if something changed
+                                _consoleWindowImpl.DirtyRegions.AddRect(new Rect(head, new Size(1, 1)));
+                                _pixelBuffer[head] =
+                                    pixel.Blend(new Pixel(moveBrush.CaretStyle));
                             }
-
-                            return;
                         }
+
+                        return;
+                    }
                 }
 
                 FillRectangleWithBrush(brush, r);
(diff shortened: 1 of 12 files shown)


src/Consolonia.Core/Drawing/PixelBufferImplementation/BoxPattern.cs
```diff
@@ -1,61 +1,27 @@
...
     /// </summary>
-    [DebuggerDisplay("DrawingBox {Text}")]
-    [JsonConverter(typeof(SymbolConverter))]
-    public readonly struct DrawingBoxSymbol : ISymbol, IEquatable<DrawingBoxSymbol>
+    public static class BoxPattern
     {
         // all 0bXXXX_0000 are special values
-        private const byte BoldSymbol = 0b0001_0000;
-        private const byte EmptySymbol = 0b0;
+        public const char Min = (char)0x2500;
```
jinek: will be less questions if you put just character values

src/Consolonia.Core/Drawing/PixelBufferImplementation/BoxPattern.cs
```diff
@@ -1,61 +1,27 @@
...
-            if (symbolAbove.NothingToDraw()) return this;
-
-            if (symbolAbove is not DrawingBoxSymbol drawingBoxSymbol)
-                return symbolAbove;
-
-            if (drawingBoxSymbol.UpRightDownLeft == BoldSymbol || UpRightDownLeft == BoldSymbol)
-                return new DrawingBoxSymbol(BoldSymbol);
+        // all 0bXXXX_0000 are special values
+        public const char BoldChar = '█';
+        public const byte BoldMask = 0b0001_0000;
```
jinek: But it's not mask, right? It's the value itself. Only one single possible value. Or is it mask? Because if it is just one value - we steal only one of 256 values. If it's mask - we still 32 values which all represent same.

src/Consolonia.Core/Drawing/PixelBufferImplementation/BoxPattern.cs
```diff
@@ -86,8 +52,8 @@ private static char GetBoxSymbol(byte upRightDownLeft)
                 {
                     return upRightDownLeft switch
                     {
-                        EmptySymbol => char.MinValue,
-                        BoldSymbol => '█',
+                        EmptyMask => char.MinValue,
```
jinek: Similar about this. How it works it does not look like a mask, it looks like a value.

src/Consolonia.Core/Drawing/PixelBufferImplementation/BoxPattern.cs
```diff
@@ -120,47 +86,27 @@ private static char GetBoxSymbol(byte upRightDownLeft)
...
                         0b0110_0111 or 0b0011_0111 or 0b0111_0111 => '╦',
-                        _ => throw new InvalidOperationException()
+                        _ => throw new InvalidOperationException(GetMaskText(upRightDownLeft))
                     };
                 }
             }
         }
 
-        public static DrawingBoxSymbol UpRightDownLeftFromPattern(byte pattern, LineStyle lineStyle)
+        public static byte Merge(byte boxMask, byte overlayMask)
```
jinek: Can we rename all these for better self explainatory? Is it merging drawing box symbols? Then may be also not the best to use *mask

src/Consolonia.Core/Drawing/PixelBufferImplementation/BoxPattern.cs
```diff
@@ -120,47 +86,27 @@ private static char GetBoxSymbol(byte upRightDownLeft)
...
-                case LineStyle.Bold:
-                    return new DrawingBoxSymbol(BoldSymbol);
-                case LineStyle.DoubleLine:
-                    byte leftPart = (byte)(pattern << 4);
-                    return new DrawingBoxSymbol((byte)(leftPart | pattern));
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(lineStyle), lineStyle, null);
-            }
-        }
+            if (overlayMask == EmptyMask)
```
jinek: Is this check redundant?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+        public Symbol(char ch)
+        {
+            Text = SymbolCache[ch];
+            if (Text == null)
+            {
+                Text = ch.ToString();
+                SymbolCache[ch] = Text;
+            }
+
+            Width = (byte)Text.MeasureText();
```
jinek: Should we cache Width as well?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                Text = string.Empty;
+            }
+            else if (boxChar >= BoxPattern.Min && boxChar <= BoxPattern.Max)
+            {
+                // get cached string instance
+                Text = SymbolCache[boxChar];
+                if (Text == null)
+                {
+                    Text = boxChar.ToString();
```
jinek: We are having some misunderstanding here, I wrote about this two times. May be I should add a test case to explain better? Let me know, please. When drawing box is created by us - it's supposed to be merged with other drawing box. But if it comes to us as a text - we must proceess it as all other regular symbols. For instance: if we draw border around textbox - that lines becomes edge characters at the edges. But if a user types drawing box symbol into that textbox - it must not interact with other drawing box.

jinek: I started to dig further - it seems more like my misunderstanding

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+        public string Text { get; }
+
+        // box pattern for box merging.
+        public byte Pattern { get; }
+
+        [JsonIgnore] public byte Width { get; init; }
+
+        public bool NothingToDraw()
+        {
+            return string.IsNullOrEmpty(Text);
```
jinek: Correct me. Text can never be null? Which means we can just do object.ReferenceEquals(string.Empty) ? Which is same is Text == string.Empty ?

jinek: Seems my comment is lost. Text can never be null

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        public Symbol Blend(ref Symbol symbolAbove)
+        {
+            if (symbolAbove.NothingToDraw())
+                return this;
+
+            if (IsBoxSymbol() && symbolAbove.IsBoxSymbol())
```
jinek, line 124: Something wrong with this logic. If one pixel is empty and another is drawing box, the result must be drawing box. Because otherwise it will stop merging from next layer.
(Lets add a test case for this)
I think, that IsBoxSymbol must be true for empty symbol. Or may be we should remove `IsBoxSymbol()` method completely and not to use this concept at all.

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+                return new Symbol(BoxPattern.Merge(Pattern, symbolAbove.Pattern));
+            return symbolAbove;
+        }
+
+        public bool IsBoxSymbol()
+        {
+            return Pattern > 0;
```
jinek, line 133: So may be it should be inclusive

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        public override bool Equals([NotNullWhen(true)] object obj)
+        {
+            return obj is Symbol other && Text.Equals(other.Text, StringComparison.Ordinal) &&
```
jinek: Can we throw an exception or do Assert for debugging to be sure it's not called?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -744,7 +744,23 @@ private void DrawBoxPixelAndMoveHead(ref Point head, Line line, LineStyle lineSt
             _consoleWindowImpl.DirtyRegions.AddRect(intersect);
         }
 
-        private void DrawLineSymbolAndMoveHead(ref Point head, bool isVertical, in ISymbol symbol, Color color,
+        public static byte UpRightDownLeftFromPattern(byte pattern, LineStyle lineStyle)
```
jinek: If moved here, lets make private then?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -744,7 +744,23 @@ private void DrawBoxPixelAndMoveHead(ref Point head, Line line, LineStyle lineSt
...
+        {
+            switch (lineStyle)
+            {
+                case LineStyle.SingleLine:
+                    return pattern;
+                case LineStyle.DoubleLine:
+                    byte leftPart = (byte)(pattern << 4);
```
jinek: I'm lost. Do you understand how it worked considering bold symbol is 0b0001_0000  ?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: I don't think so, if pattern is zero we aren't tracking any single or double lines. This makes sense to me and it's what we use for all of the chars that aren't sourced from box pattern.

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: added width cache

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: changed to Text.Length == 0; <-- there is rule which says should be used for Empty comparison.

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: This is OK
* if top is empty it keeps bottom.
* if both are box it merges them
* Default: Top symbol wins (including top being a box)

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: The next step ors them together, here's the real operation:
return (byte)((pattern << 4) | pattern); => 0b0001_0001

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: Unit tests excercise, I did verify that while running gallery it never hits

PR #441, opened 2025-09-18
further optimizations

jinek:


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -547,8 +554,21 @@ private void FillRectangleWithBrush(IBrush brush, Rect r)
                             _pixelBuffer[x, y] = _pixelBuffer[x, y].Invert();
                             break;
                         default:
+                            Pixel pixelAbove;
+                            if (solidColorBrush == null)
```
tomlm, line 558: good change, this was on my list to change. I had noticed we were calling that over and over.

PR #451, opened 2025-10-01
cherry pick of bug fixes from edit.net

tomlm:
* Implement search in AvaloniaEdit
* Fix accessText
* fix menu
* cleanup gallery 
* fix line number in avalonia edit
* storage bug fixes
* theme switching now completely switches themes by swapping mainwindow content


src/Consolonia.AvaloniaEdit/ConsoleLineNumberMargin.cs
```diff
@@ -0,0 +1,37 @@
+using System.Globalization;
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Media;
+using AvaloniaEdit.Editing;
+using AvaloniaEdit.Rendering;
+
+namespace Consolonia.AvaloniaEdit
+{
+    public class ConsoleLineNumberMargin : LineNumberMargin
```
jinek, line 10: Is the only point of this class is to create FormattedText manually? Can we just override factory for that?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -163,16 +163,20 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
+                    if (brush != null &&
+                        brush.Opacity > 0 &&
+                        streamGeometry.Fills.Count > 0)
                         foreach (Rectangle fill in streamGeometry.Fills)
                             DrawRectangle(brush, pen, new RoundedRect(fill.Rect));
 
-                    // if we have strokes to draw
-                    if (streamGeometry.Strokes.Count > 0)
+                    // if we have strokes to draw, and a valid pen 
+                    if (pen != null &&
```
jinek, line 174: It's much different logic. Do we know why before it was drawing strokes using the brush? May be there was a point to draw background under lines. If no lines - just background?

src/Consolonia.Gallery/View/ControlsListView.axaml
```diff
@@ -22,48 +21,41 @@
...
-                                <Template>
-                                    <CheckBox BorderThickness="0"
-                                              IsHitTestVisible="False"
-                                              IsChecked="{Binding $parent[MenuItem].IsChecked}" />
-                                </Template>
-                            </Setter.Value>
-                        </Setter>
+                    <Style Selector="MenuItem[IsChecked=true]">
+                        <Setter Property="Icon"
+                                Value="{StaticResource Icon_Check}" />
```
jinek: This icon does not work everywhere. Checkbox does.

src/Consolonia.AvaloniaEdit/ConsoleLineNumberMargin.cs
tomlm: this will go away with new AvaloniaEdit bug fix.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: My old code was bogus. When I got a pen and brush if I didn't have a pen I would CREATE one. This created big goofy rectangles around search hit highlights and was actually incorrect.
This code is correct, it correctly draws when it's supposed to instead of guessing like the old code did.

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: Hummm.....I guess I was assuming that we would create icons that do work everywhere, aka that these would be part of the theming. Maybe that's a bad idea?

jinek: I don't know. Idea is may be fine, but for turbo vision we must select from https://en.wikipedia.org/wiki/Code_page_437

jinek: Sorry for dirty screen. This is how icons page looks like in Linux Console

jinek: image

jinek: Or may be those icons should be used in modern themes only. That solves the issue, right?

tomlm: yeah, I guess. kind of sucks

jinek (dismissed):

jinek (approved):

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: as per teams chat, changed to using menuitem theme with turbo using diamond char.

jinek (dismissed):

jinek (dismissed):

jinek (approved):

PR #452, opened 2025-10-03
Make WheelMouse and ScrollBars scroll appropriately 

tomlm:
Works around issues in Avalonia/AvaloniaEdit tracked in bug https://github.com/AvaloniaUI/AvaloniaEdit/issues/540

* Fix up/down pageup/pagedown behavior for scroll bars being hard coded to 50/100
 by using ScrollBar theme to size to the scrollviewer viewport via new SizeConverters

* Fix WheelMouse scrolling up and down 50/100 by implementing custom ConsoleScrollContentPresenter which overrides OnWheelMouse handler with appropriate values.

* Fix misc bug found in gallery app of out of order parameters bug when calling Messagebox in GalleryEvents.
* changed gallery list to be listbox because DataGrid has scroll issues deeper than this PR
* Fix Listbox gallery tab order and unit test because it was driving me crazy the way it navigated.

### DIFF
--- a/src/Consolonia.Gallery/View/ControlsListView.axaml
+++ b/src/Consolonia.Gallery/View/ControlsListView.axaml
@@ -63,15 +63,18 @@
             </MenuItem>
         </Menu>
 
-        <DataGrid x:Name="GalleryGrid"
+        <ListBox x:Name="GalleryGrid"
                   DockPanel.Dock="Left"
                   SelectedIndex="0"
+                  Focusable="true"
+                  Width="23"
                   SelectionMode="Single">
-            <DataGrid.Columns>
-                <DataGridTextColumn Header="Name"
-                                    Binding="{Binding Name}" />
-            </DataGrid.Columns>
-        </DataGrid>
+            <ListBox.ItemTemplate>
+                <DataTemplate>
+                    <TextBlock Text="{Binding Name}" />
+                </DataTemplate>
+            </ListBox.ItemTemplate>
+        </ListBox>
 
         <Border
             Child="{Binding ElementName=GalleryGrid,Path=SelectedItem, Converter={StaticResource GalleryItemConverter}}"
--- a/src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
+++ b/src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Reflection;
+using Avalonia;
+using Avalonia.Controls.Presenters;
+using Avalonia.Controls.Primitives;
+using Avalonia.Input;
+using Avalonia.Utilities;
+
+namespace Consolonia.Themes.Templates.Controls.Helpers
+{
+    // This is workaround for Avalonia issue https://github.com/AvaloniaUI/AvaloniaEdit/issues/540
+    // Wheel mouse values are hard coded.
+    public class ConsoleScrollContentPresenter : ScrollContentPresenter
+    {
+
+        // Cache MethodInfo for private base method to avoid repeated reflection lookup.
+        private static readonly MethodInfo s_snapOffsetMethod =
+            typeof(ScrollContentPresenter).GetMethod("SnapOffset", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.FlattenHierarchy);
+
+        protected override void OnPointerWheelChanged(PointerWheelEventArgs e)
+        {
+            // base.OnPointerWheelChanged(e);
+            if (Extent.Height > Viewport.Height || Extent.Width > Viewport.Width)
+            {
+                var scrollable = Child as ILogicalScrollable;
+                var isLogical = scrollable?.IsLogicalScrollEnabled == true;
+
+                var x = Offset.X;
+                var y = Offset.Y;
+                var delta = e.Delta;
+
+                // KeyModifiers.Shift should scroll in horizontal direction. This does not work on every platform. 
+                // If Shift-Key is pressed and X is close to 0 we swap the Vector.
+                // NOTE: Changed to also include CTRL
+                if ((e.KeyModifiers == KeyModifiers.Control || e.KeyModifiers == KeyModifiers.Shift) && 
+                    MathUtilities.IsZero(delta.X))
+                {
+                    delta = new Vector(delta.Y, delta.X);
+                }
+                else
+                {
+                    delta = AdjustDeltaForFlowDirection(delta, FlowDirection);
+                }
+
+                if (Extent.Height > Viewport.Height)
+                {
+                    // double height = isLogical ? scrollable!.ScrollSize.Height : 50;
+                    double height = 3;
+                    y += -delta.Y * height;
+                    y = Math.Max(y, 0);
+                    y = Math.Min(y, Extent.Height - Viewport.Height);
+                }
+
+                if (Extent.Width > Viewport.Width)
+                {
+                    //double width = isLogical ? scrollable!.ScrollSize.Width : 50;
+                    double width = 3;
+                    x += -delta.X * width;
+                    x = Math.Max(x, 0);
+                    x = Math.Min(x, Extent.Width - Viewport.Width);
+                }
+
+                var newOffset = (Vector)s_snapOffsetMethod.Invoke(this, new object[] { new Vector(x, y), delta, true })!;
+
+                bool offsetChanged = newOffset != Offset;
+                SetCurrentValue(OffsetProperty, newOffset);
+
+                e.Handled = !IsScrollChainingEnabled || offsetChanged;
+            }
+        }
+        private static Vector AdjustDeltaForFlowDirection(Vector delta, Avalonia.Media.FlowDirection flowDirection)
+        {
+            if (flowDirection == Avalonia.Media.FlowDirection.RightToLeft)
+            {
+                return delta.WithX(-delta.X);
+            }
+            return delta;
+        }
+
+    }
+}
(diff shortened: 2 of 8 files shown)


src/Consolonia.Gallery/View/ControlsListView.axaml
```diff
@@ -63,15 +63,18 @@
...
-                  SelectionMode="Single">
-            <DataGrid.Columns>
-                <DataGridTextColumn Header="Name"
-                                    Binding="{Binding Name}" />
-            </DataGrid.Columns>
-        </DataGrid>
+        <ListBox x:Name="GalleryGrid"
+                 DockPanel.Dock="Left"
+                 SelectedIndex="0"
+                 Focusable="true"
```
jinek: Isn't `true` a default value?

src/Consolonia.Gallery/View/ControlsListView.axaml
```diff
@@ -63,15 +63,18 @@
...
+                 Width="24"
```
jinek: Why do we need fixed width?

src/Consolonia.Gallery/View/ControlsListView.axaml
```diff
@@ -63,15 +63,18 @@
...
+                 SelectionMode="Single">
+            <ListBox.ItemTemplate>
+                <DataTemplate>
```
jinek, line 78: This is probably default template of the ListBox?

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Reflection;
+using Avalonia;
+using Avalonia.Controls.Presenters;
+using Avalonia.Input;
+using Avalonia.Media;
+using Avalonia.Utilities;
+
+namespace Consolonia.Themes.Templates.Controls.Helpers
+{
+    // This is workaround for Avalonia issue https://github.com/AvaloniaUI/AvaloniaEdit/issues/540
```
jinek, line 12: How to reproduce this issue? When I scroll with the wheel in edit.net branch it seems all good.

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
```diff
@@ -0,0 +1,74 @@
...
+        private static readonly MethodInfo SnapOffsetMethod =
+            typeof(ScrollContentPresenter).GetMethod("SnapOffset",
+                BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.FlattenHierarchy);
+
+        protected override void OnPointerWheelChanged(PointerWheelEventArgs e)
+        {
+            // base.OnPointerWheelChanged(e);
+            if (Extent.Height > Viewport.Height || Extent.Width > Viewport.Width)
+            {
+                // var scrollable = Child as ILogicalScrollable;
```
jinek: But this looks like breaking cases when child implements own scrolling.

src/Consolonia.Themes/Templates/Controls/ScrollBar.axaml
```diff
@@ -169,10 +172,18 @@
         <Style Selector="^:vertical">
             <Setter Property="Width"
                     Value="1" />
+            <Setter Property="SmallChange"
+                    Value="1" />
+            <Setter Property="LargeChange"
```
jinek: It's not for sure this ScrollBar is used by a scroll viewer upper in the tree, It looks like this value should be set in a style in a ScrollViewer theme. Or may be directly on the scrollbar itself?

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: The listbox is in a dock panel and the listbox width changes as you scroll. I don't know if that's an avalonia issue or not, but when you scroll down to TransitioningContent the dockpanel gets bigger.

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: Um. I need to bind to Name. How do I do that without an ItemTemplate?

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
tomlm: You "fixed" it in curses by hacking the wheel scroll value to hide it. It's actually scrollling 50 pixels but you are scrolling on 1/12 with wheel mouse. This ends up scrolling 4.16 pixels. 

On windows we use scroll wheel amount of 1.  Avalonia normally uses a value like 1. See the delta in curses

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
tomlm: yeah, I backed off. I defer to child unless it's AvaloniaEdit (since we know it has a bug.)

src/Consolonia.Themes/Templates/Controls/ScrollBar.axaml
tomlm: good call, I moved this up to the scrollviewer

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: removed

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: Image

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
tomlm: Change the curses velocity to one
```   const double velocity = 1;```

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: This is because default is virtualizingStackPanel. This is avalonia behavior. 
I fixed this in this case by use stackpanel instead of virtualizing panel so width is computed on all items.

src/Consolonia.Gallery/View/ControlsListView.axaml
tomlm: I don't understand why, but selection logic for unit tests all fail if we don't set this as focusable. Something about the unit test is not setting the focus correctly unless I do this.

src/Consolonia.Gallery/View/ControlsListView.axaml
jinek: DisplayMemberBinding property

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
jinek: understood. What if we just change velocity to match the needed scroll speed?

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
jinek: Is this Presenter now needed only for AvaloniaEdit? Because if so, it would make sense implement all this stuff right inside AvaloniaEdit project.

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
tomlm: More information:
Avalonia off the shelf generates wheel events with velocity delta = -1, 0, 1
I surveyed all of the avalonia code base and AvaloniaEdit and this is what I found:
The following controls implement OnWheel()  uses positive/negative delta for next/previous:
* **buttonSpinner** , **Combo** , **Calendar** , **CalendarDatePicker** 

The following controls implement OnWheel() event handler but takes delta and multiples by 50
* **DataGrid** 

The following control implements IlogicalScrollable 
* **AvaloniaEdit** -> default ScrollSize => x=10, y=50

**ScrollViewer** uses ScrollContentPresenter in it's template
* **ScrollContentPresenter** -> uses ILogicalScrollable.ScrollSize (but defaults to 50 pixels) * delta

**ItemsControl** uses ItemsPresenter
* **ItemsPresenter** -> Simple defers to Panel as ILogicalScrollable

The following panels have ILogicalScrollable implemented
* **VirtualizingCarouselPanel** - returns ScrollSize of (1,1) and PageScrollSize of (1,1)
* **DateTimePickerPanel** - returns ScrollSize of (0,ItemHeight) and PageScrollSize of (0, ItemHeight*4)

My delta was predicated on following thinking:
* Consolonia should behave like Avalonia when generating events so delta should be -1, 0, 1, not doing this has potential of breaking code that is expecting those values.
* Introducing ConsoleScrollContentPresenter via ScrollViewer's template which changes the default scroll size for controls that aren't ILogicalScrollable effectively fixes all controls without custom paging. 
* Changing ConsoleScrollContentPresenter to detect AvaloniaEdit broken ILogicalScrollable works as short term hack until they fix their code. 
* DataGrid - Not certain what to do with this. It doesn't use ScrollViewer AFAIK and it is completely beyond this delta. It does look like it implements it's own OnPointerWheel() event to scroll, and so in theory we could create a ConsoleDataGrid which overrides this, but (shrug)?

src/Consolonia.Themes/Templates/Controls/Helpers/ConsoleScrollContentPresenter.cs
tomlm: see long response on previous issue. Yes, it fixes any control which doesn't have a ILogicalScrollable implementation.

PR #453, opened 2025-10-03
Use AvaloniaEdit for ViewCode dialog

tomlm:
Add AvaloniaEdit and hook up the things.
Fix bug in template for slider
avaloniaedit

### DIFF
--- a/src/Consolonia.Gallery/Gallery/GalleryViews/GalleryEvents.axaml.cs
+++ b/src/Consolonia.Gallery/Gallery/GalleryViews/GalleryEvents.axaml.cs
@@ -28,14 +28,14 @@ namespace Consolonia.Gallery.Gallery.GalleryViews
         private void OnRawMouse(RawPointerEventType type, Point point, Vector? nullable, RawInputModifiers modifiers)
         {
             var eventViewModel = new RawMouseEventViewModel(type, point, nullable, modifiers);
-            ViewModel.RawMouseEvents.Add(eventViewModel);
+            ViewModel.AddRawMouseEvent(eventViewModel);
             ViewModel.SelectedRawMouseEvent = eventViewModel;
         }
 
         private void OnRawKey(Key arg1, char arg2, RawInputModifiers arg3, bool arg4, ulong arg5, bool arg6)
         {
             var eventViewModel = new RawKeyboardEventViewModel(arg1, arg2, arg3, arg4, arg5, arg6);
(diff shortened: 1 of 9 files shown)


src/Consolonia.Gallery/Consolonia.Gallery.csproj
```diff
@@ -12,13 +12,18 @@
     </ItemGroup>
     
     <ItemGroup>
+        <ProjectReference Include="..\Consolonia.AvaloniaEdit\Consolonia.AvaloniaEdit.csproj" />
```
jinek, line 15: Lets discuss this. If we move AvaloniaEdit to a separate repository, it can become an inconsistency for AvaloniaEdit beacuse it will assume one version of Consolonia, why current can be newer. Same time this project is used for automatic testing. Thus results become dependent on that.

src/Consolonia.Themes/Templates/Controls/Slider.axaml
```diff
@@ -137,9 +137,9 @@
             <Setter Property="Ticks"
                     Value="{TemplateBinding Ticks}" />
         </Style>
-        <Style Selector="^:disabled /template/ Grid#grid">
+        <!--<Style Selector="^:disabled /template/ Grid#grid">
```
jinek, line 140: ) nice

src/Consolonia.Gallery/Consolonia.Gallery.csproj
tomlm: I don't think I understand what you are saying. Can you explain more?

jinek: Once we introduced organization and split projects to several repos, this would become cross reference between repos. And it would be inconsistient dependency on versions. Plus execution of e2e tests against Gallery would become dependent on one more external tool. Which probably should have tests itself in own repo.

tomlm: ah, I was thinking Consolonia.AvaloniaEdit is part of core libraries. Technically Avalonia edit is **Avalonia.AvaloniaEdit** package, and so clearer part of Avalonia.

jinek: As I understand same in Avalonia. Separate repo, separate package. If I install Avalonia it does not bring AvaloniaEdit. Thoughts?

PR #454, opened 2025-10-03
Clipping on line symbols was incorrect when the line went negative

tomlm:
causing it to draw too many symbols shifted to the right.

Added unit tests

image

### DIFF
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -750,15 +750,12 @@ namespace Consolonia.Core.Drawing
             Rect rectToRefresh = line.Vertical
                 ? new Rect(head.X, head.Y, 1, count)
                 : new Rect(head.X, head.Y, count, 1);
-            Rect intersect = CurrentClip.Intersect(rectToRefresh);
-            if (intersect.IsEmpty())
-                return;
 
-            double start = line.Vertical ? intersect.Top : intersect.Left;
-            double end = line.Vertical ? intersect.Bottom : intersect.Right;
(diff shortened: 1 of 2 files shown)


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -750,12 +750,9 @@ private void DrawBoxPixelAndMoveHead(ref Point head, Line line, LineStyle lineSt
             Rect rectToRefresh = line.Vertical
                 ? new Rect(head.X, head.Y, 1, count)
                 : new Rect(head.X, head.Y, count, 1);
-            Rect intersect = CurrentClip.Intersect(rectToRefresh);
-            if (intersect.IsEmpty())
-                return;
```
jinek, line 755: Why do we need to do anything outside of the clipping? Do we know which commit brought the regression?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -788,31 +760,54 @@ private static byte GetBoxPatternFromLineStyle(byte pattern, LineStyle lineStyle
...
-                    ? head.WithY(head.Y + 1)
-                    : head.WithX(head.X + 1);
+                if (isVertical)
+                    y++;
+                else
+                    x++;
             }
 
-            _consoleWindowImpl.DirtyRegions.AddRect(intersect);
+            _consoleWindowImpl.DirtyRegions.AddRect(CurrentClip.Intersect(lineBounds));
```
jinek, line 810: You are already heaving the result of intersection in `intersectLine`

tomlm: I'm fixing in my next PR

PR #455, opened 2025-10-09
Change to use PixelPoint/PixelRect for clipping and as intermediate format between Point/Rect and PixelBufferCoordinate/PixelBuffer

tomlm:
Avalonia has 2 data structures
* **Point/Rect** - double based
* **PixelPoint/PixelRect** - int based

When we are dealing with clipping and mapping between double based point/rect and the underlying **PixelBuffer** we need to make sure that we are being precise about the translation between the worlds. For example:
* casting a negative double point truncates TOWARDs zero, so we need to use Math.Floor()
* Managing Clipping and DirectRects should be with integer precision, not double precision.

This PR implements this and is explicit about that mapping, DrawingContextImpl methods should consistently do the following:
1. incoming Point or Rect is in doubles
2. We translate the Point or Rect to apply origin transforms to new Point/Rect
3. We Convert from virtual double based **Point/Rect** to pixel oriented int based **PixelPoint/PixelRect**. We use PixelPoint/PixelRect for all of our internal layout logic.
4. When interacting with PixelBuffer we cast the final **PixelPoint/PixelRect** to **PixelBufferCoordinate** which is ushort based. 

Essentially, all internal logic is PixelPoint/PixelRect based so that we are dealing with pixel calculations including boundaries that can be negative outside of clipping regions etc. and then PixelBufferCoordinate is specifically for interacting with the PixelBuffer.

In the process of doing this I updated the tests and in the process of updating the tests I discovered that **Rectangle** with a stroke is different than a Border around a rectangle. The issue is that a geometry with Stroke is modeled differently in Avalonia  

* **Border around a rectangle** in Avalonia is always "outside" the rectangle it's wrapping. This is what we use 99% of the time and so all of our logic around that is correct and battle tested
image

* **Geometry with a stroke** in Avalonia is modeled as half over lay of the virtual line of the outside of the geometry (aka Rectangle). Avalonia will muck with the rectangle and positioning to achieve this by offsetting the rectangle by Thickness/2 (aka 1=>.5) and making the width and height be Thickness less. In a pixel mapping system this approximation is subtle and you would never really see it, but in Consolonia making the rect 1 less in each dimension is 1 char less in both directions. And we can't draw a line at a half overlaid position. 

Because of this issue I spent a considerable amount of time trying to come up with a reasonable policy.  Strokes essentially are meant to be overlaid on top of the rectangle, so that you can do things like dashed lines that have the background color come through. This spirtually works with our LineStyle of Single/Double lines, because the line is drawn on the colored background giving us a line drawn "inside" the rectangle like this: 
image
As opposed to a policy where we draw single line outside of the rectangle, which would look like this:
image

To do this I need to restore the size of the rectangle to the original size and then draw the single/doubleline on the edges of that rectangle.

But when we use Edge line styles, the line has to be drawn outside of the rectangle (because there are no chars for the insides corners, they are made up of 2 outside chars abutting). So to support edge line styles I need to resize the rectangle and then draw the lines outside that resized rectangle, like this 
image

I currently only do this when inside **DrawGeometry()** with Rectangle comes in with a Brush and with a Pen

I have not been able to hit any instances where this happens via DrawRectangle() directly, if that exists we will have to do something more invasive. I also don't know if StreamGeometry which comes in as collection of Fills and Strokes independent of each other has this issue or not. I don't know that there is anything I can with it at the moment, so I'm going to pretend that lives the realm of "don't do that". :) 

I added a bunch of unit tests to validate this with much improved messages so you can see what the F*ck happened. 

I also updated Pixel/Symbol/BoxPattern to display their values in the debugger so I didn't claw my eyeballs out trying to figure out what box pattern was laid down in the background.
Here is an example: 
image
This shows that the symbol is horizontal line, and the underlying box pattern is Single Left and Single Right
image

And this one shows the end of the line is a horizontal line char, but the underlying box pattern is only single left (SR).

### DIFF
--- a/src/Consolonia.Themes/Templates/Controls/Slider.axaml
+++ b/src/Consolonia.Themes/Templates/Controls/Slider.axaml
@@ -139,7 +139,7 @@
         </Style>
         <Style Selector="^:disabled /template/ Grid#grid">
             <Setter Property="Opacity"
-                    Value="{DynamicResource ThemeNoDisturbBrush}" />
+                    Value=".8" />
         </Style>
     </ControlTheme>
 </ResourceDictionary>


src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBufferCoordinate.cs
```diff
@@ -25,6 +25,20 @@ public static explicit operator PixelBufferCoordinate((ushort x, ushort y) val)
 
         public static explicit operator PixelBufferCoordinate(Point point)
         {
+#if DEBUG
```
jinek: Why not to throw it in !DEBUG ? Otherwise we are just having wrong coordinates.

tomlm: I guess I was thinking this is hot path and I want to know it's wrong while developing, not runtime. I got rid of #if debug

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
```diff
@@ -70,9 +70,16 @@ public Pixel this[Point point]
             set => this[(PixelBufferCoordinate)point] = value;
```
jinek: This indexer now can be removed (based on your PR description)

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBufferCoordinate.cs
jinek: I agree with you. It looks like a hot path. What if rather than throwing by ourselves we just disable math overflows on the project level?

jinek: But actually this is struct, comparison will be executed by CPU registries only. just thoughts.. I don't know. Let's leave //todo: replace this check by math overflow at project level. Fine for you?

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
tomlm: yeah, I removed this from PixelBuffer, I also removed Point cast operator on PixelBufferCoordinate. You have to do ushort or PixelPoint.

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBufferCoordinate.cs
tomlm: added todo

PR #460, opened 2025-10-11
Custom edge handling  in DrawGeometry also applies to DrawRectangle

tomlm:
As suspected, this actually does apply to DrawRectangle. This happens when you have a Border with a background brush. I didn't see it because Modern theme doesn't do this, but Turbo does.

Changes:
* Moved custom edge logic from DrawGeometry to DrawRectangle. DrawGeometry just calls DrawRectangle
* Changed policy to shrink fill for edge lines
* Changed DrawGeometry test to be DrawRectangle instead of DrawGeometry, this unit test covers both cases.
* Changed DrawBox unit test to not have background brush as that's not what it's testing, we don't want to activate the brush/pen to test raw box drawing logic.

image

### DIFF
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -226,7 +226,7 @@ namespace Consolonia.Core.Drawing
                 ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.DrawingRoundedOrNonUniformRectandle, this,
                     brush, pen, roundedRect, boxShadows);
 
-            var rect = roundedRect.Rect;
+            Rect rect = roundedRect.Rect;
             if (pen != null && brush != null)
             {
                 // This is one of those places where Avalonia/Consolonia don't align well due to character nature of consolonia.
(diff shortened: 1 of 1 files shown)


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -239,84 +215,50 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
+            if (roundedRect.Rect.IsEmpty()) return;
 
-            if (rect.Rect.IsEmpty()) return;
+            if (roundedRect.IsRounded)
+                ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.DrawingRoundedOrNonUniformRectandle, this,
+                    brush, pen, roundedRect, boxShadows);
 
-            if (rect.IsRounded)
+            Rect rect = roundedRect.Rect;
+            if (pen == null)
```
jinek: This two IFs can be merged to `if(pen == null || brush == null)`. I'm not sure it's better tho. Resolve this youself please, it's just a note

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -239,84 +215,50 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
-                                _pixelBuffer[head] =
-                                    pixel.Blend(new Pixel(moveBrush.CaretStyle));
-                            }
-                        }
-
-                        return;
-                    }
+                    DrawRectangleInternal(brush, null,
+                        new Rect(rect.Position.X + 1, rect.Position.Y + 1, rect.Width - 1, rect.Height - 1));
+                    DrawRectangleInternal(null, pen, rect, boxShadows);
```
jinek: second call of DrawRectangle with pen is same in both branches. You can just move it after if/else.

PR #461, opened 2025-10-11
Upgrade avalonia to 11.3.7

tomlm:
* Upgrade to avalonia 11.3.7

Implement new clipboard model.  Platforms no longer implement **IClipboard** and instead implement the simpler **IClipboardImpl** interface. There is an internal **Clipboard** class which consumes platform IClipboardImpl.
 * Renamed **InprocessClipboard** to **ConsoleClipboard** which is base class which implements new IClipboardImpl interface by deferring to simply SetTextAsync and GetTextAsync. 
 * Changed all of the clipboard implmentations to derive from ConsoleClipboard base class and override GetText/SetText methods since we only support clipboard operations externally via text. 
 * Changed instantiation to use relfection to create Clipboard wrapper around IClipboardImpl
 * Updated unit tests etc. 

This new implementation should correctly do complex clipboard opeartions in process and only reflect the clipboard state to external platform for text operations.

### DIFF
--- a/src/Consolonia.Core/Infrastructure/AsyncDataTransfer.cs
+++ b/src/Consolonia.Core/Infrastructure/AsyncDataTransfer.cs
@@ -8,7 +8,7 @@ namespace Consolonia.Core.Infrastructure
 {
     public sealed class AsyncDataTransfer : IAsyncDataTransfer
     {
-        private List<IAsyncDataTransferItem> _items;
+        private readonly List<IAsyncDataTransferItem> _items;
         private bool _disposedValue;
 
         public AsyncDataTransfer()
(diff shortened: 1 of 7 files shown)


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -188,37 +188,37 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
-                    if (pen != null &&
-                        pen.Thickness > 0 &&
-                        pen.Brush != null &&
-                        pen.Brush.Opacity > 0 &&
-                        streamGeometry.Strokes.Count > 0)
                     {
-                        RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
-                        for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
+                        // if we have fills to do and a brush with opacity
+                        if (brush != null &&
```
jinek: Is this part of upgrade?

tomlm: no, I started doing it on the old branch.  Now that everything is checked in and merged it's just version change and redoing the clipboard stuff

PR #462, opened 2025-10-13
Fix Render artifacts

tomlm:
artifacts

When we were doing cursor and dragging there were a number of artifacts (turds) left behind. I dug into it and there were 3 issues:
* dirtyrect for cursor doesn't take into account that unicode sequences may need to render outside of dirty rect.
* unicode library which measures text was saying the menu hamburger in windows width was 1 when it should be 2.
* WindowsManager library didn't set cursor when over title bar, making it look like resize icon was a glitch.
* When moving cursor over wide chars we didn't refresh/redraw correctly, leaving invalidated turds behind that would go away when full screen refresh.

Fixes are:
* Upgrade wcwidth library to 3.0, fixes measuretext bug
* Updated WindowsManager library which fixes cursor ontitle bar
* Ripped out convulated and complicated empty pixel mechanism 
  * Moved logic for what to actually draw into RenderTarget. 
  * Expanded DirtyRect left and right to make sure that complex chars are rendered for cursor operations.
  * Instead of manipulating PixelBuffer to clear out hidden chars I set cache to Empty and use expanded dirty rects

The new algo is:
**DrawaContextImpl**
* WideChars get written as a pixel just like all pixels.
* DirtyRects for DrawString is now the full string clipped, but expanded to make sure that prior and after widechars are rendered.

**RenderTarget** 
* Rendering a wide char invalidates the cache for covered chars with empty, if empty pixels are part of dirtyrect than "real" pixels from pixelbuffer will be rendered.
* If cursor overlaps a pixel (including wide pixels), then the cursor is rendered instead, clearing the underlying cache for the pixels the cursor covers so that future dirtyrect can render them correctly.

Added a bunch of unit tests around new policy.

This greatly simplifies the logic of dealing with wide chars, by making it so that any code writing a symbol into PixelBuffer doesn't have to know fancy rules for how to do that. It's all centralized in the Rendering/cache layer.

### DIFF
--- a/src/Consolonia.Core/Drawing/RenderTarget.cs
+++ b/src/Consolonia.Core/Drawing/RenderTarget.cs
@@ -97,8 +97,8 @@ namespace Consolonia.Core.Drawing
 
             // initialize the cache with Pixel.Empty as it literally means nothing
             for (ushort y = 0; y < height; y++)
-                for (ushort x = 0; x < width; x++)
-                    cache[x, y] = Pixel.Empty;
+            for (ushort x = 0; x < width; x++)
+                cache[x, y] = Pixel.Empty;
 
             return cache;
         }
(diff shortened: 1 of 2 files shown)


src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -878,61 +878,37 @@ private void DrawLineSymbolAndMoveHead(ref PixelPoint head, bool isVertical, in
                     case "\r":
                     case "\f":
                     case "\n":
-                        position = new PixelPoint(lineStartX, position.Y + 1);
+                        position = new PixelPoint(startPosition.X, position.Y + 1);
                         break;
                     default:
```
tomlm, line 882: this is the meat of the change. I got rid of all of this complex pixelbuffer manipulation. The model is now super simple, pixels get written to the buffer. and the dirty rect is as appropriate for the string written.,

src/Consolonia.Core/Infrastructure/ConsoleCursor.cs
```diff
@@ -1,13 +1,23 @@
...
 {
-    public readonly struct ConsoleCursor(PixelBufferCoordinate coordinate, string type)
+    public readonly struct ConsoleCursor
         : IComparable<ConsoleCursor>, IEquatable<ConsoleCursor>
     {
-        public PixelBufferCoordinate Coordinate { get; init; } = coordinate;
-        public string Type { get; init; } = type;
+        public PixelBufferCoordinate Coordinate { get; init; }
+        public string Type { get; init; }
+        public byte Width { get; init; }
```
tomlm, line 12: I added width to consolecursor as optimization, we use this value over and over.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -188,37 +188,37 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
-                    // if we have strokes to draw, and a valid pen 
-                    if (pen != null &&
-                        pen.Thickness > 0 &&
-                        pen.Brush != null &&
-                        pen.Brush.Opacity > 0 &&
-                        streamGeometry.Strokes.Count > 0)
                     {
-                        RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
-                        for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
+                        // if we have fills to do and a brush with opacity
```
tomlm: no change here, I don't know why the diff is showing a difference.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -276,28 +276,28 @@ public void DrawGeometry(IBrush brush, IPen pen, IGeometryImpl geometry)
...
+                            return;
+                        }
                     case MoveConsoleCaretToPositionBrush moveBrush:
-                    {
-                        var head = rectangleRect.TopLeft.ToPixelPoint();
-                        if (CurrentClip.ContainsExclusive(head))
                         {
-                            Pixel pixel = _pixelBuffer[head];
-                            if (pixel.CaretStyle != moveBrush.CaretStyle)
+                            var head = rectangleRect.TopLeft.ToPixelPoint();
```
tomlm: only get the pixel if it's in the rect

jinek (approved):

Issue #477, opened 2025-10-16
When switching from "Icons" to "Image" and back in gallery some artifacts appear


jinek:
When switching from "Icons" to "Image" and back in gallery some artifacts appear
<img width="1494" height="1028" alt="image" src="https://github.com/user-attachments/assets/d6cc1f6c-d301-4a10-9b8d-d27eac16dcb5" />

_Originally posted by @jinek in https://github.com/jinek/Consolonia/issues/458#issuecomment-3411182649_

tomlm:
this is same issue, do a clean build

PR #481, opened 2025-10-21
Add support for Text and Emoji variation annotations

tomlm:
This is to make Consolonia more robust for dealing with complex emoji surrogates. issue: #477 

# Issue
some characters are ambiguous as to whether they are rendered as single size or double size.  For compatibility reasons most terminals default to an ambiguous undefined behavior. (See https://github.com/microsoft/terminal/issues/16852).  To give authors more control over interpretation of ambiguous sequences unicode/terminals support Variation selectors
* **Text Variation (U+FE0E)** - previous character should be interpreted as text, which usually means single wide
* **Emoji Variation (U+FE0F)** - previous character should be interpreted as emoji, which usually means double wide

In addition, some terminals will advance the cursor 2 characters when rendering wide chars, and some (Gnome ubuntu terminal) advance the cursor always one character. 

# Changes
* Changes **MeasureText()** to honor Text/Emoji variation when calculating widths
* Changes **GetGlyphs()** to pass through Text/Emoji variations so that Glyphs contain them
* Changes **Symbol()** to keep variation if it's passed, but if it's not, it will automatically add the Emojii Variation to characters we know are Emojii's or think should be wide.
* Changes **static Icons** so that if they are all rendered as double chars, either because they are already recognized emojii's, or because we have added the Emoji Variation to the icon static template.
* Added detection of fallback (on linux) for SupportsEmojiVariations. Linux renders them differently but only advances 1 char.  When SupportsEmojiVariations == false then we explicitly advance the cursor for wide chars to make it match our avalonia rendering layout.
* Added handling for Skin variations, keycap and regional annotations.
* New glyph unit tests 
* Updated unit tests 

# Effects
This has the effect of:
* Defaults for emojiis are always wide, an we pass that on to the terminal so it can know to render it wide.
* If a character doesn't render the way you want, you can add the variation selector to get it do that.
```cs
   var wide = $"🠷\uFE0F"; 
   var narrow = $"🠷\uFE0E"; 
```
```xaml
   <TextBlock>🠷&#xFE0E; Wide</TextBlock>
   <TextBlock>🠷&#xFE0E; Narrow</TextBlock>
```
Icons before changes, showing how chars overlap with following character:
image

Icons after changes:
image

### DIFF
--- a/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
+++ b/src/Consolonia.Core/Drawing/DrawingContextImpl.cs
@@ -164,38 +164,38 @@ namespace Consolonia.Core.Drawing
                     DrawLineInternal(pen, myLine);
                     break;
                 case StreamGeometryImpl streamGeometry:
-                    {
-                        // if we have fills to do and a brush with opacity
-                        if (brush != null &&
-                            brush.Opacity > 0 &&
-                            streamGeometry.Fills.Count > 0)
-                            foreach (Rectangle fill in streamGeometry.Fills)
-                                // Investigate: Does the pen apply to rectangle or not?
-                                DrawRectangle(brush, null, new RoundedRect(fill.Rect));
+                {
+                    // if we have fills to do and a brush with opacity
+                    if (brush != null &&
+                        brush.Opacity > 0 &&
+                        streamGeometry.Fills.Count > 0)
+                        foreach (Rectangle fill in streamGeometry.Fills)
+                            // Investigate: Does the pen apply to rectangle or not?
+                            DrawRectangle(brush, null, new RoundedRect(fill.Rect));
 
-                        // if we have strokes to draw, and a valid pen 
-                        if (pen != null &&
-                            pen.Thickness > 0 &&
-                            pen.Brush != null &&
-                            pen.Brush.Opacity > 0 &&
-                            streamGeometry.Strokes.Count > 0)
+                    // if we have strokes to draw, and a valid pen 
+                    if (pen != null &&
+                        pen.Thickness > 0 &&
+                        pen.Brush != null &&
+                        pen.Brush.Opacity > 0 &&
+                        streamGeometry.Strokes.Count > 0)
+                    {
+                        RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
+                        for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
                         {
-                            RectangleLinePosition[] strokePositions = InferStrokePositions(streamGeometry);
-                            for (int iStroke = 0; iStroke < streamGeometry.Strokes.Count; iStroke++)
-                            {
-                                Line stroke = streamGeometry.Strokes[iStroke];
-                                RectangleLinePosition strokePosition = strokePositions[iStroke];
-                                if (strokePosition == RectangleLinePosition.Left)
-                                    DrawLineInternal(pen, stroke, RectangleLinePosition.Left);
-                                else if (strokePosition == RectangleLinePosition.Right)
-                                    DrawLineInternal(pen, stroke, RectangleLinePosition.Right);
-                                else if (strokePosition == RectangleLinePosition.Top)
-                                    DrawLineInternal(pen, stroke, RectangleLinePosition.Top);
-                                else if (strokePosition == RectangleLinePosition.Bottom)
-                                    DrawLineInternal(pen, stroke, RectangleLinePosition.Bottom);
-                            }
+                            Line stroke = streamGeometry.Strokes[iStroke];
+                            RectangleLinePosition strokePosition = strokePositions[iStroke];
+                            if (strokePosition == RectangleLinePosition.Left)
+                                DrawLineInternal(pen, stroke, RectangleLinePosition.Left);
+                            else if (strokePosition == RectangleLinePosition.Right)
+                                DrawLineInternal(pen, stroke, RectangleLinePosition.Right);
+                            else if (strokePosition == RectangleLinePosition.Top)
+                                DrawLineInternal(pen, stroke, RectangleLinePosition.Top);
+                            else if (strokePosition == RectangleLinePosition.Bottom)
+                                DrawLineInternal(pen, stroke, RectangleLinePosition.Bottom);
                         }
                     }
+                }
                     break;
                 default:
                     ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.DrawGeometryNotSupported, this, brush,
@@ -436,28 +436,28 @@ namespace Consolonia.Core.Drawing
                     case VisualBrush:
                         throw new NotImplementedException();
                     case ISceneBrush sceneBrush:
-                        {
-                            ISceneBrushContent sceneBrushContent = sceneBrush.CreateContent();
-                            sceneBrushContent?.Render(this, Matrix.Identity);
-                            return;
-                        }
+                    {
+                        ISceneBrushContent sceneBrushContent = sceneBrush.CreateContent();
+                        sceneBrushContent?.Render(this, Matrix.Identity);
+                        return;
+                    }
                     case MoveConsoleCaretToPositionBrush moveBrush:
+                    {
+                        var head = rectangleRect.TopLeft.ToPixelPoint();
+                        if (CurrentClip.ContainsExclusive(head))
                         {
-                            var head = rectangleRect.TopLeft.ToPixelPoint();
-                            if (CurrentClip.ContainsExclusive(head))
+                            Pixel pixel = _pixelBuffer[head];
+                            if (pixel.CaretStyle != moveBrush.CaretStyle)
                             {
-                                Pixel pixel = _pixelBuffer[head];
-                                if (pixel.CaretStyle != moveBrush.CaretStyle)
-                                {
-                                    // only be dirty if something changed
-                                    _consoleWindowImpl.DirtyRegions.AddRect(new PixelRect(head, new PixelSize(1, 1)));
-                                    _pixelBuffer[head] =
-                                        pixel.Blend(new Pixel(moveBrush.CaretStyle));
-                                }
+                                // only be dirty if something changed
+                                _consoleWindowImpl.DirtyRegions.AddRect(new PixelRect(head, new PixelSize(1, 1)));
+                                _pixelBuffer[head] =
+                                    pixel.Blend(new Pixel(moveBrush.CaretStyle));
                             }
-
-                            return;
                         }
+
+                        return;
+                    }
                 }
 
                 FillRectangleWithBrush(brush, rectangleRect.ToPixelRect());
@@ -616,18 +616,18 @@ namespace Consolonia.Core.Drawing
             {
                 case ShadeBrush:
                     for (ushort y = (ushort)targetRect.Y; y < targetRect.Bottom; y++, brushY++)
-                        for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
-                            _pixelBuffer[x, y] = _pixelBuffer[x, y].Shade();
+                    for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
+                        _pixelBuffer[x, y] = _pixelBuffer[x, y].Shade();
                     break;
                 case BrightenBrush:
                     for (ushort y = (ushort)targetRect.Y; y < targetRect.Bottom; y++, brushY++)
-                        for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
-                            _pixelBuffer[x, y] = _pixelBuffer[x, y].Brighten();
+                    for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
+                        _pixelBuffer[x, y] = _pixelBuffer[x, y].Brighten();
                     break;
                 case InvertBrush:
                     for (ushort y = (ushort)targetRect.Y; y < targetRect.Bottom; y++, brushY++)
-                        for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
-                            _pixelBuffer[x, y] = _pixelBuffer[x, y].Invert();
+                    for (ushort x = (ushort)targetRect.X; x < targetRect.Right; x++)
+                        _pixelBuffer[x, y] = _pixelBuffer[x, y].Invert();
                     break;
                 default:
                     for (ushort y = (ushort)targetRect.Y; y < (ushort)targetRect.Bottom; y++, brushY++)
--- a/src/Tests/Consolonia.Core.Tests/WithLifetimeFixture/DrawingContextImplTests.cs
+++ b/src/Tests/Consolonia.Core.Tests/WithLifetimeFixture/DrawingContextImplTests.cs
@@ -181,7 +181,7 @@ namespace Consolonia.Core.Tests.WithLifetimeFixture
             var dc = new DrawingContextImpl(consoleTopLevelImpl);
             DrawText(dc, 0, 0, "A\tB", Brushes.White);
             Assert.IsTrue(buffer[0, 0].Foreground.Symbol.Character == 'A');
-            for (ushort i =1; i < 5; i++)
+            for (ushort i = 1; i < 5; i++)
                 Assert.IsTrue(buffer[i, 0].Foreground.Symbol.Character == ' ');
             Assert.IsTrue(buffer[5, 0].Foreground.Symbol.Character == 'B');
         }
--- a/src/Tests/Consolonia.Core.Tests/WithLifetimeFixture/GlyphTests.cs
+++ b/src/Tests/Consolonia.Core.Tests/WithLifetimeFixture/GlyphTests.cs
@@ -1,4 +1,5 @@
 //DUPFINDER_ignore
+
 using System.Collections.Generic;
 using Consolonia.Controls;
 using Consolonia.Core.Drawing.PixelBufferImplementation;


jinek:
I'm not sure if it's supposed to be fixed. Attaching as a non-blocking comment 

https://github.com/user-attachments/assets/65e84f7e-8b0a-45f9-a73c-a5380731e193

src/Consolonia.Controls/IConsoleCapabilities.cs
```diff
@@ -6,5 +6,7 @@ public interface IConsoleCapabilities
         ///     This is true if console supports composing multiple emojis together (like: 👨‍👩‍👧‍👦).
         /// </summary>
         bool SupportsComplexEmoji { get; }
+
+        bool SupportsEmojiVariation { get; }
```
jinek, line 10: If this is always false when first property is false, then better to create 3-states enum

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
```diff
@@ -86,15 +86,21 @@ public string PrintBuffer()
...
-
-                    //todo: check why cursor is not drawing
-                    stringBuilder.Append(text);
+                    if (pixel.Width > 0)
+                    {
+                        string text = pixel.IsCaret() ? "Ꮖ" : pixel.Foreground.Symbol.GetText();
+
+                        //todo: check why cursor is not drawing
+                        stringBuilder.Append(text);
+                        i += pixel.Width;
```
jinek, line 102: What would be width for Caret?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -33,6 +42,20 @@ public Symbol(char ch)
             Complex = null;
             Width = (byte)UnicodeCalculator.GetWidth(ch);
             Pattern = 0;
+            // if we think it should be wide, OR we know it's an emoji 
+            if (Width == 2 || Emoji.IsEmoji(new string(ch, 1)))
+            {
+                // we want to use EmojiVariation to signal we think it's wide.
+                Character = char.MinValue;
+                Width = 2;
+                lock (GlyphCharCache)
```
jinek, line 51: I bet ConcurrentDictionary would work faster. Just my guess.

src/Consolonia.Core/Text/TextShaper.cs
```diff
@@ -11,6 +12,9 @@ namespace Consolonia.Core.Text
 {
     public class TextShaper : ITextShaperImpl
     {
+        private static readonly Dictionary<ushort, string> GlyphByIndex = new();
+        private static readonly Dictionary<string, ushort> IndexByGlyph = new();
+
```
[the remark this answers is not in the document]
tomlm: fixed

src/Consolonia.Core/Text/TextShaper.cs
```diff
@@ -20,12 +24,35 @@ public ShapedBuffer ShapeText(ReadOnlyMemory<char> text, TextShaperOptions optio
...
             return shapedBuffer;
         }
+
+        public static string GetGlyphByIndex(ushort glyphIndex)
+        {
+            lock (IndexByGlyph)
+            {
+                return GlyphByIndex[glyphIndex];
+            }
+        }
```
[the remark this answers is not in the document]
tomlm: fixed

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -74,12 +76,40 @@ public void Print(PixelBufferCoordinate bufferPoint, Color background, Color for
...
+                // rendering over the top with the glyph.
+                // process each glyph, rendering the width as spaces then moving the cursor and
+                foreach (string glyph in str.GetGlyphs(SupportsComplexEmoji))
+                {
+                    ushort glyphWidth = glyph.MeasureText();
+                    if (glyphWidth > 1)
+                    {
+                        WriteText(Esc.SetCursorPosition(bufferPoint.X, bufferPoint.Y));
+                        WriteText(new string(' ', Math.Min(Size.Width - bufferPoint.X, bufferPoint.X + glyphWidth)));
+                    }
```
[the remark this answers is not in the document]
tomlm: fixed

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -74,12 +76,40 @@ public void Print(PixelBufferCoordinate bufferPoint, Color background, Color for
...
             WriteText(sb.ToString());
 
             ushort textWidth = str.MeasureText();
+
+            // move to position
+            if (SupportsEmojiVariation)
+            {
+                SetCaretPosition(bufferPoint);
+                WriteText(str);
+            }
```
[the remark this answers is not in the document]
tomlm: fixed

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -74,12 +76,40 @@ public void Print(PixelBufferCoordinate bufferPoint, Color background, Color for
...
+            {
+                // rendering over the top with the glyph.
+                // process each glyph, rendering the width as spaces then moving the cursor and
+                foreach (string glyph in str.GetGlyphs(SupportsComplexEmoji))
+                {
+                    ushort glyphWidth = glyph.MeasureText();
+                    if (glyphWidth > 1)
+                    {
+                        WriteText(Esc.SetCursorPosition(bufferPoint.X + 1, bufferPoint.Y));
+                        WriteText(new string(' ', Math.Min(Size.Width - bufferPoint.X, glyphWidth - 1)));
```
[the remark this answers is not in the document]
tomlm: already fixed

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
tomlm: The width of the caret comes from the symbol used, so this should be good.

src/Consolonia.Controls/IConsoleCapabilities.cs
tomlm: I think they can be separate. One honors control markers for width/rendering, the other is about chaining multiple glyphs like skin color.

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: image
image

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
```diff
@@ -86,15 +87,19 @@ public string PrintBuffer()
...
                     Pixel pixel = this[new PixelBufferCoordinate(i, j)];
-                    string text = pixel.IsCaret() ? "Ꮖ" : pixel.Foreground.Symbol.GetText();
-
-                    //todo: check why cursor is not drawing
-                    stringBuilder.Append(text);
+                    if (pixel.Width > 0)
+                    {
+                        string text = pixel.IsCaret() ? "Ꮖ" : pixel.Foreground.Symbol.GetText();
+                        stringBuilder.Append(text);
+                        i += text.MeasureText();
```
jinek: Why don't we just add pixel.Width? Why are we measuring text again?

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
```diff
@@ -86,15 +87,19 @@ public string PrintBuffer()
...
-                    if (i == Width - 1 && j == Height - 1)
-                        break;
                     Pixel pixel = this[new PixelBufferCoordinate(i, j)];
-                    string text = pixel.IsCaret() ? "Ꮖ" : pixel.Foreground.Symbol.GetText();
-
-                    //todo: check why cursor is not drawing
-                    stringBuilder.Append(text);
+                    if (pixel.Width > 0)
+                    {
+                        string text = pixel.IsCaret() ? "Ꮖ" : pixel.Foreground.Symbol.GetText();
```
jinek: Why do we assume Caret pixel is always more than 0 width?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -45,22 +68,49 @@ public Symbol(byte boxPattern)
...
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public Symbol(string glyph)
+        public Symbol(string glyph, byte? width = null)
         {
-            if (glyph.Length == 1)
+            Pattern = 0;
+            Complex = null;
+            Character = char.MinValue;
+            if (string.IsNullOrEmpty(glyph))
```
jinek: What is the case glyph is null? I feel it's misconception that null and empty are same. Why do we need null glyph at all?

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
```diff
@@ -45,22 +68,49 @@ public Symbol(byte boxPattern)
...
+                    {
+                        if (GlyphComplexCache.TryGetValue(glyph, out string complex))
+                        {
+                            Complex = complex;
+                        }
+                        else
+                        {
+                            // use text variation for narrow glyphs, emoji variation for wide glyphs
+                            if (Width == 1)
+                                Complex = GlyphComplexCache[glyph] = $"{glyph}{TextVariation}";
```
jinek: It's almost same string, you can simplify to ..{width==1?... : ...}

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -277,9 +275,63 @@ public void DrawGlyphRun(IBrush foreground, IGlyphRunImpl glyphRun)
                     return;
             }
 
-            var shapedBuffer = (ShapedBuffer)glyphRunImpl.GlyphInfos;
-            string text = shapedBuffer.Text.ToString();
-            DrawStringInternal(foreground, text, glyphRun.GlyphTypeface);
+            if (foreground is not ISolidColorBrush solidColorBrush)
```
jinek, line 281: Can we avoid inlining here? Coz it's very difficult to compare the code

src/Consolonia.Core/Helpers/InputProcessing/PasteBlockMatcher.cs
```diff
@@ -2,6 +2,6 @@
 
 namespace Consolonia.Core.Helpers.InputProcessing
 {
-    public class PasteBlockMatcher<T>(Action<string> onComplete, Func<T, char> toChar)
-        : StartsEndsWithMatcher<T>(onComplete, toChar, @"[200~", @"[201~");
+    public class PasteBlockMatcher<T>(Action<string> onComplete, Func<T, string> toText)
```
jinek: Should be Rune rather than string? I thought why do we need char or string anyway? or otherwise char1,chat2 could be better?

src/Consolonia.Core/Infrastructure/DefaultNetConsoleOutput.cs
```diff
@@ -111,6 +115,19 @@ public virtual void PrepareConsole()
             _originalForeground = Console.ForegroundColor;
             _originalBackground = Console.BackgroundColor;
             Size = new PixelBufferSize((ushort)Console.WindowWidth, (ushort)Console.WindowHeight);
+            // Detect complex emoji support by writing a complex emoji and checking cursor position.
+            // If the cursor moves 2 positions, it indicates proper rendering of composite surrogate pairs.
+            (int left, _) = Console.GetCursorPosition();
+            WriteText("👨‍👩‍👧‍👦");
```
jinek, line 121: i guess for this printing we should set foreground to be same as background. This way user will not see this internal this easy

src/Consolonia.Core/Text/ConsoleTypeface.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// </summary>
+    public sealed class ConsoleTypeface : IGlyphTypeface
+    {
+        private static readonly object GlyphCacheSync = new();
+        private static readonly Dictionary<ushort, string> GlyphTextByIndex = new();
+        private static readonly Dictionary<ushort, ushort> GlyphWidthByIndex = new();
+        private static readonly Dictionary<string, ushort> GlyphIndexByText = new();
+
+        public void Dispose()
+        {
```
jinek, line 20: can clear dictionaries btw

src/Consolonia.Themes/Modern/Controls/MenuItem.axaml
```diff
@@ -13,7 +13,7 @@
             <Style Selector="^[IsChecked=true]">
                 <!-- ReSharper disable Xaml.StaticResourceNotResolved -->
                 <Setter Property="Icon"
-                        Value="{StaticResource Icon_Check}" />
+                        Value="✓" />
```
jinek, line 16: why this replace?

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
```diff
@@ -74,21 +75,39 @@ public static AppBuilder UseAutoDetectClipboard(this AppBuilder builder)
...
-                    clipboardImpl = new X11Clipboard();
-                else
-                    clipboardImpl = new ConsoleClipboard();
+                try
+                {
+
+                    if (IsWslPlatform())
+                        clipboardImpl = new WslClipboard();
+                    else
+                        clipboardImpl = new X11Clipboard();
```
jinek: If there is no DISPLAY variable, no need to try to create X11Clipboard. But also no need to try XClip as it works on top of X11.

tomlm: xclip does work on top of wayland too Does Display work for that?

src/Consolonia.Themes/Modern/Controls/MenuItem.axaml
tomlm: Well. (sigh). Most of the Icons are double wide icons. In the menu a double wide icon looks kind of silly. So I now view Icons as definitions of double wide icons, since they include things like EmojiVariation annotations etc to make them consistently icon-ish. 

In the modern theme, I didn't really like the fat checkmark there, the single char checkmark looks more "modern" to me, but it's not really a double-wide "ICON". 

That was my thinking at any rate.
image
image

src/Consolonia.Core/Helpers/InputProcessing/PasteBlockMatcher.cs
tomlm: ToUTF16() might be more accurate name.

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
tomlm: good point.I had debug code in GetText()  that changed the text but that is not normal.

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
tomlm: I can't think of circumstances where we have a caret that's not aligned on a pixel with content that has width.

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: we don't. Empty is useful for situations where we are blending partial pixel properties

src/Consolonia.Core/Drawing/PixelBufferImplementation/Symbol.cs
tomlm: ok

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: ok, but the code got rewritten a fair bit, it's still not apples to apples.  We changed from character based logic to glyph based logic.

src/Consolonia.Core/Helpers/InputProcessing/PasteBlockMatcher.cs
tomlm: It feels a bit weird to have to pass a callback for this at all.

src/Consolonia.Core/Infrastructure/DefaultNetConsoleOutput.cs
tomlm: I tried that, but foreground is ignored for complex emojii rendering. If we can find one which is multicodepoint that renders using text color than  I agree.

src/Consolonia.Core/Text/ConsoleTypeface.cs
tomlm: ok

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
jinek: no, there is another variable to detect wayland (probably `WAYLAND_DISPLAY`).
So how is it?
if DISPLAY => x11 => X11ClipBoard
otherwise try XCLIP and if fails - fallback to internal clipboard?

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
jinek: we have constructor which allows setting caret for any pixels https://github.com/jinek/Consolonia/blob/8bebc40039e8ad34e9d52e43eeb00c0eb25fda62/src/Consolonia.Core/Drawing/PixelBufferImplementation/Pixel.cs#L87
Which means we can create a pixel with width 0 and Caret. Is it correct?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
jinek: I see some changes, like parameters were removed from `ConsoloniaPlatform.RaiseNotSupported`. I can't figure why, is that by mistake?

src/Consolonia.Core/Helpers/InputProcessing/PasteBlockMatcher.cs
jinek: no, I mean different. UTF32 can be still passed over the stack. While strings will be always operated from the heap. I'm just thinking, that if we need two chars, lets replace it with two chars or whatever types is enough. Because string gives possibility to store many bytes which looks like we don't need and same time it bring the heap to operation. On other hand why to bother - it's just input. you decide to resolve this thread please

tomlm: I ended up going with Func<T, Rune>(). It turns out that a single char.ConvertToUTF16() internally simply does a new Rune(cp). Append, AppendLine, etc all will accept a rune, so it makes it clear that we are dealing with a rune.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -797,6 +819,37 @@ private static byte GetBoxPatternFromLineStyle(byte pattern, LineStyle lineStyle
             }
         }
 
+        private void DrawGlyphInfoInternal(Color foregroundColor, GlyphInfo glyphInfo, ConsoleTypeface glyphTypeface,
+            PixelPoint position)
+        {
+            // NOTE: we clip at the position of the wide char. If we attempt to clip for the width of the wide
+            // char it introduces artifacts when a wide char is partially clipped.
+            if (CurrentClip.ContainsExclusive(position))
```
jinek: If start position is clipped we still might want to render other parts which are not clipped. I guess this if should be moved to inside "else" block

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
jinek: I have reviewed this code. Another one thread was opened. this one I close as not actual anymore.

src/Consolonia.Core/Drawing/PixelBufferImplementation/PixelBuffer.cs
tomlm: I updated.

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: fixed

PR #498, opened 2025-11-06
fix keyboard navigation for storage dialogs

tomlm:


src/Consolonia.Core/Controls/FileOpenPicker.axaml.cs
```diff
@@ -1,15 +1,19 @@
...
 using Avalonia.Input;
 using Avalonia.Interactivity;
 using Avalonia.Platform.Storage;
+using Avalonia.Threading;
 using Iciclecreek.Avalonia.WindowManager;
 
 namespace Consolonia.Core.Controls
 {
-    internal partial class FileOpenPicker : ManagedWindow
+    internal partial class FileOpenPicker : ManagedWindow, IDisposable
```
jinek: But Dispose is never called for this dialogs https://github.com/jinek/Consolonia/blob/be82d91a71ed8968bd23e2200eaea827ea4a55ed/src/Consolonia.Core/Infrastructure/ConsoloniaStorageProvider.cs#L27

src/Consolonia.Core/Controls/FileOpenPicker.axaml.cs
```diff
@@ -85,8 +127,18 @@ private void ListBox_SelectionChanged(object sender, SelectionChangedEventArgs e
...
                         ViewModel.SelectedFiles.Remove(file);
             }
+        }
 
-            ViewModel.HasSelection = ViewModel.SelectedFiles.Count > 0;
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing) ItemsListBox.Items.CollectionChanged -= Items_CollectionChanged;
```
jinek: This unsubsrcribe seems not needed because both objects will be collected together

src/Consolonia.Core/Controls/FileOpenPickerViewModel.cs
```diff
@@ -9,9 +10,8 @@ namespace Consolonia.Core.Controls
 {
     internal partial class FileOpenPickerViewModel : PickerViewModelBase<FilePickerOpenOptions>
     {
-        [ObservableProperty] private bool _hasSelection;
-
-        [ObservableProperty] private ObservableCollection<IStorageFile> _selectedFiles = new();
+        [NotifyPropertyChangedFor(nameof(HasSelection))] [ObservableProperty]
```
jinek: This field is set only once and then instance never changes. Both attributes don't do anything

src/Consolonia.Core/Controls/FolderPickerViewModel.cs
```diff
@@ -8,9 +9,8 @@ namespace Consolonia.Core.Controls
 {
     internal partial class FolderPickerViewModel : PickerViewModelBase<FolderPickerOpenOptions>
     {
-        [ObservableProperty] private bool _hasSelection;
-
-        [ObservableProperty] private ObservableCollection<IStorageFolder> _selectedFolders = new();
+        [NotifyPropertyChangedFor(nameof(HasSelection))] [ObservableProperty]
+        private ObservableCollection<IStorageFolder> _selectedFolders = new();
```
jinek: Seems same: redundant attributes

src/Consolonia.Core/Controls/ListBoxExtensions.cs
```diff
@@ -0,0 +1,20 @@
+using Avalonia.Controls;
+
+namespace Consolonia.Core.Controls
+{
+    internal static class ListBoxExtensions
+    {
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```
jinek: Why do you need nullable reference type if it's not allowed in this project anyway?

tomlm: because all of the linter stuff starts bitching about setting things to null.

jinek: at the pipeline you mean? could you please demonstrate - I then will fix lint better

src/Consolonia.Core/Controls/FolderPickerViewModel.cs
tomlm: cleaned up attributes

src/Consolonia.Core/Controls/FileOpenPickerViewModel.cs
tomlm: cleaned up attributes

src/Consolonia.Core/Controls/FileOpenPicker.axaml.cs
tomlm: got rid of disposable pattern and detaching

src/Consolonia.Core/Controls/FileOpenPicker.axaml.cs
tomlm: got rid of disposer

jinek (approved):

PR #535, opened 2025-11-19
Remove Skia as core dependency

tomlm:
This delta removes Skia as core dependency. It turns out the only thing we use Skia for is for bitmap support.
This change does the following:
* Removed Skia as dependency from all Consolonia libraries.
* Changed .UseConsolonia() to detect if there is a platfomrender registered (aka Skia) and wraps it.  Essentially, If you add Avalonia.Skia and .UseSkia() then we will use fallback to the bitmap implementation that Skia registers with avalonia to create and manipulate bitmaps. (aka CreateBitmap is handled by Skia, and returns the implentation they already have as IBitmapImpl)
* Rewrites the quad pixel logic to be built using Avalonia abstract bitmap concepts (IBitmapImpl) instead of raw Skia code.
* Added Unsupported code for when there isn't a IPlatformRenderInterface to defer to for bitmap operations.

### DIFF
--- a/src/Consolonia.Gallery/Gallery/GalleryViews/GalleryBorders.axaml
+++ b/src/Consolonia.Gallery/Gallery/GalleryViews/GalleryBorders.axaml
@@ -86,20 +86,6 @@
                                Background="AntiqueWhite" />
                 </Border>
             </WrapPanel>
-            <!--            <WrapPanel>
-                <Border BorderThickness="1">
-                    <Border.BorderBrush>
-                        <console:LineBrush LineStyle="Edge"
-                                           Brush="Gray" />
-                    </Border.BorderBrush>
-                    <Panel Background="AntiqueWhite">
-                        <Image Source="avares://Consolonia.Gallery/Resources/happy.png"
-                               Width="40"
-                               Height="20" />
-                    </Panel>
-                </Border>
-
-            </WrapPanel>
--->        </StackPanel>
+        </StackPanel>
     </ScrollViewer>
 </UserControl>


src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+    internal partial class DrawingContextImpl
+    {
+        public void DrawBitmap(IBitmapImpl source, double opacity, Rect sourceRect, Rect destRect)
+        {
+            if (source is DummyBitmap)
+                return;
+
+            // resize bitmap to destination rect size
+            var targetRect = new Rect(Transform.Transform(new Point(destRect.TopLeft.X, destRect.TopLeft.Y)),
+                    Transform.Transform(new Point(destRect.BottomRight.X, destRect.BottomRight.Y)))
```
jinek: TopLeft and BottomRight are already of type Point

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            // Get the platform render interface to resize the bitmap
+            var renderInterface = AvaloniaLocator.Current.GetRequiredService<IPlatformRenderInterface>();
+
+            // Resize source to be target rect * 2 so we can map to quad pixels
+            var targetSize = new PixelSize(targetRect.Width * 2, targetRect.Height * 2);
+            using IBitmapImpl resizedBitmap =
+                renderInterface.ResizeBitmap(source, targetSize, BitmapInterpolationMode.MediumQuality);
+
+            // Get pixel data from the resized bitmap
```
jinek: These comments are no use..

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+            var readableBitmap = (IReadableBitmapImpl)resizedBitmap;
+
+            using ILockedFramebuffer frameBuffer = readableBitmap.Lock();
+
+            int stride = frameBuffer.RowBytes;
+            int bytesPerPixel = frameBuffer.Format.BitsPerPixel / 8;
+            byte[] pixelBytes = new byte[stride * frameBuffer.Size.Height];
+            Marshal.Copy(frameBuffer.Address, pixelBytes, 0, pixelBytes.Length);
```
jinek: You don't need to copy the frame each time. You can create new Span<byte>((void*)frameBuffer.Address, length)) and then cast it using MemoryMarshal.Cast like you did

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            int py = targetRect.TopLeft.Y;
+
+            for (int y = 0; y < targetSize.Height; y += 2, py++)
+            {
+                int px = targetRect.TopLeft.X;
+                for (int x = 0; x < targetSize.Width; x += 2, px++)
+                {
+                    var point = new PixelPoint(px, py);
+                    if (CurrentClip.ContainsExclusive(point))
```
jinek: We can intersect outside of the loops and run loops only thru intersection. If you don't want to do it in this task (it seems out of scope) then drop a comment please or even better a ticket.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+using Avalonia.Media;
+using Avalonia.Media.Imaging;
+using Avalonia.Platform;
+using Consolonia.Core.Drawing.PixelBufferImplementation;
+using Consolonia.Core.Dummy;
+
+namespace Consolonia.Core.Drawing
+{
+    [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    internal struct BgraColor
```
jinek: can be readonly? May be we could take advantage, like by passing by stack reference?

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            _consoleWindowImpl.DirtyRegions.AddRect(targetRect);
+        }
+
+        public void DrawBitmap(IBitmapImpl source, IBrush opacityMask, Rect opacityMaskRect, Rect destRect)
+        {
+            throw new NotImplementedException();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jinek: I've just tried to attach the debugger to RELEASE compiled run and it looks like this method is inlined by JIT automatically, this attribute does not change anything.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+        private static BgraColor GetPixelColor(Span<BgraColor> pixels, int x, int y, int stride, int bytesPerPixel)
```
jinek: I've tried to run benchmark to test inlining:

```
Inlining:
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13588.834 ms, avg=1.3589 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13510.921 ms, avg=1.3511 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=10074.252 ms, avg=1.0074 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=7310.767 ms, avg=0.7311 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13632.462 ms, avg=1.3632 ms



No Inlining:
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13426.475 ms, avg=1.3426 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13451.543 ms, avg=1.3452 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13583.190 ms, avg=1.3583 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13508.435 ms, avg=1.3508 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=11971.686 ms, avg=1.1972 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=12915.532 ms, avg=1.2916 ms

No Attribute:
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13561.013 ms, avg=1.3561 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=5864.973 ms, avg=0.5865 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=11444.361 ms, avg=1.1444 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13144.431 ms, avg=1.3144 ms
[DEBUG_LOG] DrawBitmap: iterations=10000, total=13562.994 ms, avg=1.3563 ms
```

thus I demad benchmark for AggressiveInlining or any other value use.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+            using ILockedFramebuffer frameBuffer = readableBitmap.Lock();
+
+            int stride = frameBuffer.RowBytes;
+            int bytesPerPixel = frameBuffer.Format.BitsPerPixel / 8;
+            byte[] pixelBytes = new byte[stride * frameBuffer.Size.Height];
+            Marshal.Copy(frameBuffer.Address, pixelBytes, 0, pixelBytes.Length);
+            Span<BgraColor> pixels = MemoryMarshal.Cast<byte, BgraColor>(pixelBytes);
+
+            // this is reused by each pixel as we process the bitmap
+            Span<BgraColor> quadPixelColors = stackalloc BgraColor[4];
```
jinek: Why do we need this pixels? It looks like a sub-array. Can't we create a span directly with a shift from original array?

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            for (int y = 0; y < targetSize.Height; y += 2, py++)
+            {
+                int px = targetRect.TopLeft.X;
+                for (int x = 0; x < targetSize.Width; x += 2, px++)
+                {
+                    var point = new PixelPoint(px, py);
+                    if (CurrentClip.ContainsExclusive(point))
+                    {
+                        // get the quad pixel from the bitmap as a quad of 4 BgraColor values
```
jinek: Honestly I don't understand the comment

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+        {
+            char character = GetColorsPattern(colors) switch
+            {
+                // ReSharper disable StringLiteralTypo
+                0b0000 => ' ',
+                0b1000 => '▘',
+                0b0100 => '▝',
+                0b0010 => '▖',
+                0b0001 => '▗',
+                0b1001 => '▚',
```
jinek: Leave a todo please. Half of these characters are out of IBM code page. It looks bad where not supported.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+                throw new ArgumentException($"{nameof(pixelColors)} must have 4 elements.");
+
+            BgraColor bgraColor = quadPixel switch
+            {
+                ' ' => BgraColor.Transparent,
+                '▘' => pixelColors[0],
+                '▝' => pixelColors[1],
+                '▖' => pixelColors[2],
+                '▗' => pixelColors[3],
+                '▚' => CombineColors(stackalloc BgraColor[] { pixelColors[0], pixelColors[2] }),
```
jinek: Can be simplified to `[pixelColors[0], pixelColors[2]]`

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+            return new BgraColor(b, g, r, a);
+        }
+
+        /// <summary>
+        ///     Cluster quad colors into a pattern (like: TTFF) based on relative closeness
+        /// </summary>
+        /// <param name="colors"></param>
+        /// <returns>T or F for each color as a string</returns>
+        /// <exception cref="ArgumentException"></exception>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jinek: Same, I want benchmark please

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+                pixel.A = 255;
+
+            return pixel;
+        }
+
+
+        /// <summary>
+        ///     given 4 colors return quadPixel character which is suitable to represent the colors
+        /// </summary>
+        /// <param name="colors"></param>
```
jinek: THese 3 empty lines can be removed

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        private static char GetQuadPixelCharacter(Span<BgraColor> colors)
+        {
+            char character = GetColorsPattern(colors) switch
+            {
+                // ReSharper disable StringLiteralTypo
```
jinek: I don't see resharper to complaing about typos here

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+{
+    [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    internal struct BgraColor
+    {
+        public byte B;
+        public byte G;
+        public byte R;
+        public byte A;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jinek: I bet it's inlined automatically, same for other methods

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
```diff
@@ -0,0 +1,395 @@
...
+                ((clusters[0] == higherCluster ? 0b1000 : 0) |
+                 (clusters[1] == higherCluster ? 0b0100 : 0) |
+                 (clusters[2] == higherCluster ? 0b0010 : 0) |
+                 (clusters[3] == higherCluster ? 0b0001 : 0));
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool ColorEquals(BgraColor c1, BgraColor c2)
+        {
+            return c1.B == c2.B && c1.G == c2.G && c1.R == c2.R && c1.A == c2.A;
```
jinek: gemini proposed this: `return Unsafe.As<BgraColor, int>(ref c1) == Unsafe.As<BgraColor, int>(ref c2);`

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
```diff
@@ -395,712 +197,10 @@ public void PopLayer()
             ConsoloniaPlatform.RaiseNotSupported(NotSupportedRequestCode.PushLayerNotSupported);
         }
 
-        private void DrawRectangleInternal(IBrush brush, IPen pen, Rect rect, BoxShadows boxShadows = new())
```
jinek, line 398: It's not good you split the code and introduce functional changes same time. I don't know what to review - what was changed, what was moved which means I have to review everything

src/Consolonia.Core/Drawing/DrawingContextImpl.Lines.cs
```diff
@@ -0,0 +1,616 @@
...
+using System.Linq;
+using Avalonia;
+using Avalonia.Media;
+using Avalonia.Platform;
+using Consolonia.Controls.Brushes;
+using Consolonia.Core.Drawing.PixelBufferImplementation;
+using Consolonia.Core.Infrastructure;
+using Consolonia.Core.InternalHelpers;
+
+namespace Consolonia.Core.Drawing
```
jinek: Was this changes? Do I need to review?

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -180,7 +183,9 @@ private void RenderToDevice()
                     _cache[x++, y] = Pixel.Empty;
             }
 
-            flushingBuffer.Flush();
+            _flushingBuffer.Flush();
+
+            _console.EndRender();
```
jinek: I feel there is something wrong here. FlushingBuffer was introduced specifically to accumulate output to write all at once. But now this role is somehow moved to EndRender

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: you are right, I'm a bastard.  I will unsplit the code and do the split in seperate delta.

src/Consolonia.Core/Dummy/DummyBitmap.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using Avalonia;
+using Avalonia.Platform;
+
+namespace Consolonia.Core.Dummy
+{
+    // copy of Avalonia.HeadlessServer HeadlessBitmapStub with minor changes
+    internal class DummyBitmap : IDrawingContextLayerImpl, IWriteableBitmapImpl
```
jinek, line 10: As I understand we don't use this anyhow, can be all throw NotSupported?

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: fixed

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: removed some, kept some that help me

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -24,10 +24,12 @@ internal class RenderTarget : IDrawingContextLayerImpl
         // cache of pixels written so we can ignore them if unchanged.
         private Pixel?[,] _cache;
         private ConsoleCursor _consoleCursor;
+        private FlushingBuffer _flushingBuffer;
```
jinek: Do we move FlushingBuffer to field to save on object allocation? This does not give anything: FlushingBuffer is a struct itself, while BufferBuffer flushing just drops the reference to string data, which goes to GC same way.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: that introduces an unsafe, this way avoids the unsafe keyword.  Doesn't using unsafe make it so that our assembly has to be marked as unsafe?

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -180,7 +183,9 @@ private void RenderToDevice()
                     _cache[x++, y] = Pixel.Empty;
```
jinek, line 180: This X coordinate calculation is now out of sync with WritePixel. Also, In WritePixel we should advance y + 1 in case end of the line hit. Because if this and that calculations are out of sync we do unecessary Caret moves.

src/Consolonia.Core/Drawing/RenderTarget.cs
jinek: We don't need to introduce one more buffer, we can fix existing one. The redundant calls to Flush can be made to depend on the console or something. Like for AnsiConsole we can skip flushing on Caret changes, for DefaultNetConsole we do flush on Caret moves etc.

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: fixed

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: added

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: removed

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: removed

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
tomlm: sure, why not?

src/Consolonia.Core/Drawing/DrawingContextImpl.cs
tomlm: undid this, I'm sorry :(

src/Consolonia.Core/Drawing/DrawingContextImpl.Lines.cs
tomlm: not changed

src/Consolonia.Core/Dummy/DummyBitmap.cs
tomlm: no, this allows us to continue if not throw, graceful fallback to non-render bitmap.

src/Consolonia.Core/Drawing/RenderTarget.cs
tomlm: my understanding of StringBuilder  is that it will maintain a memory buffer internally, even if you hit .Clear().  So in theory instead of allocating memory to build the string, we are maintaining one allocated builder with it's internal buffer which is used to build up the output over and over.  Let me do some research and see if that's right.

tomlm: yeah:
```
Setting the Capacity and Length
Although the [StringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder) is a dynamic object that allows you to expand the number of characters in the string that it encapsulates, you can specify a value for the maximum number of characters that it can hold. This value is called the capacity of the object and should not be confused with the length of the string that the current [StringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder) holds. For example, you might create a new instance of the [StringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder) class with the string "Hello", which has a length of 5, and you might specify that the object has a maximum capacity of 25. When you modify the [StringBuilder](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder), it does not reallocate size for itself until the capacity is reached. When this occurs, the new space is allocated automatically and the capacity is doubled. 
```

src/Consolonia.Core/Drawing/DrawingContextImpl.Bitmap.cs
jinek: As far as I know it does nothing. Also you already did `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>`

src/Consolonia.Core/Drawing/RenderTarget.cs
jinek: As I see none of these means it's better to clear than re-create.

src/Consolonia.Core/Drawing/ConsoloniaRenderInterface.cs
```diff
@@ -198,7 +273,9 @@ public IPlatformRenderInterfaceContext CreateBackendContext(IPlatformGraphicsCon
 
         public bool IsSupportedBitmapPixelFormat(PixelFormat format)
         {
-            throw new NotImplementedException();
+            if (_fallback != null)
+                return _fallback.IsSupportedBitmapPixelFormat(format);
+            return false;
```
jinek, line 278: Missed return RaiseNotSupported(... nameof(IsSupportedBitmapPixelFormat)) ? I don't understand what is method is for tho

src/Consolonia.Gallery/Gallery/GalleryViews/GalleryBorders.axaml
```diff
@@ -86,7 +86,7 @@
                                Background="AntiqueWhite" />
                 </Border>
             </WrapPanel>
-            <WrapPanel>
+            <!--            <WrapPanel>
```
jinek: why?

src/Consolonia.Core/Dummy/DummyBitmap.cs
jinek: But we do return; for DummyBitmap. we are not using it anyhow, or do we?

src/Consolonia.Core/Drawing/ConsoloniaRenderInterface.cs
tomlm: I decided I didn't really need to raise an exception on this, the default of false is fine. Otherwise I have to plumb everything through and really it will have been hit already on CreateBitmap().

src/Consolonia.Gallery/Gallery/GalleryViews/GalleryBorders.axaml
tomlm: because it has a  UnitTest and would require the unit test to have dependency on Skia. It wasn't important part of borders, but I should just delete instead of commenting out.

src/Consolonia.Core/Dummy/DummyBitmap.cs
tomlm: yeah we do return it as the result for NotSupported code

src/Consolonia.Gallery/Gallery/GalleryViews/GalleryBorders.axaml
tomlm: I removed the commented markup

PR #537, opened 2025-11-21
Refactored buffering:

tomlm:
* Added Flush() method to IConsoleOutput/IConsole
* Changed RenderTarget to remove FlushBuffer and simply call Flush() at end of render pass.
* DefaultNetConsoleOutput - Buffers on textruns (essentially FlushBuffer logic merged into DefaultNetConsoleOutput) Goal is minimize API calls by consolidating consecutive pixels)
* AnsiConsoleOutput - Buffers for entire render pass (only writes to output on Flush at end of Render) 
*Added NoStrikethrough/NoItalic/NoUnderline esc codes so we don't use RESET anymore.
* Renamed Print() => WritePixel(coord, in pixel)

### DIFF
--- a/src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
+++ b/src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
@@ -93,7 +93,7 @@ namespace Consolonia.Core.Infrastructure
                 _lastTextDecoration = pixel.Foreground.TextDecoration;
             }
 
-            var style = pixel.Foreground.Style ?? FontStyle.Normal;
+            FontStyle style = pixel.Foreground.Style ?? FontStyle.Normal;
             if (style != _lastStyle)
             {
                 //reset previous style
@@ -111,14 +111,14 @@ namespace Consolonia.Core.Infrastructure
                 _lastStyle = style;
             }
 
-            var weight = pixel.Foreground.Weight ?? FontWeight.Normal;
+            FontWeight weight = pixel.Foreground.Weight ?? FontWeight.Normal;
             if (weight != _lastWeight)
             {
                 WriteText(weight switch
                 {
                     FontWeight.Bold or FontWeight.SemiBold or FontWeight.ExtraBold or FontWeight.Black =>
                         Esc.Bold,
-                    FontWeight.Thin or FontWeight.ExtraLight or FontWeight.Light => 
+                    FontWeight.Thin or FontWeight.ExtraLight or FontWeight.Light =>
                         Esc.Dim,
                     _ => Esc.Normal
                 });


src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -62,31 +74,80 @@ public void Print(PixelBufferCoordinate bufferPoint, Color background, Color for
...
+            if (style != _lastStyle)
+            {
+                //reset previous style
+                WriteText(_lastStyle switch
+                {
+                    FontStyle.Italic => Esc.NoItalic,
+                    _ => string.Empty
+                });
+
+                WriteText(style switch
```
jinek, line 106: These two `WriteText` can be replaced with just one: 

```csharp
WriteText(style switch
                {
                    FontStyle.Italic => Esc.Italic,
                    _ => Esc.NoItalic
                });
```

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -130,23 +201,24 @@ public void PrepareConsole()
...
             _supportsComplexEmoji = left2 - left == 2;
 
             // write out a char with wide variation selector
-            WriteText("\U0001F5D1\uFE0F"); // 🗑 Wastebasket + emoji variation selector
+            Console.Write("\U0001F5D1\uFE0F"); // 🗑 Wastebasket + emoji variation selector
             (int left3, _) = Console.GetCursorPosition();
             _supportsEmojiVariation = left3 - left2 == 2;
 
             ClearScreen();
+            Flush();
```
jinek, line 229: Why do you flush here? I see you used Console.Write directly

src/Consolonia.Core/Infrastructure/DefaultNetConsoleOutput.cs
```diff
@@ -51,23 +61,54 @@ public virtual PixelBufferCoordinate GetCaretPosition()
...
+            Color foreground,
             FontStyle? style,
-            FontWeight? weight, TextDecorationLocation? textDecoration, string str)
+            FontWeight? weight,
+            TextDecorationLocation? textDecoration,
+            string str)
         {
-            ConsoleColor originalForeground = Console.ForegroundColor;
-            ConsoleColor originalBackground = Console.BackgroundColor;
+            if (bufferPoint != _currentBufferPoint || _stringBuilder.Length == 0)
```
jinek: This seems wrong: Why do we move the caret if buffer is empty?

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -62,31 +74,80 @@ public void Print(PixelBufferCoordinate bufferPoint, Color background, Color for
...
+                    _lastBackground = background;
+                }
+            }
 
             // move to position
             if (SupportsEmojiVariation)
             {
                 SetCaretPosition(bufferPoint);
                 WriteText(str);
+                ushort textWidth = str.MeasureText();
```
jinek: Should not we rename Print to PrintPixel which would accept a pixel. Width would be already calculated there. Also would be no need to enumerate attributes in method parameters. Also class fields could be replaced with `_lastPixel`

src/Consolonia.Core/Text/Esc.cs
```diff
@@ -20,8 +20,11 @@ public static class Esc
 
         // text decorations
         public const string Italic = "\u001b[3m";
+        public const string NoItalic = "\u001b[23m";
```
jinek, line 23: 👍

src/Consolonia.PlatformSupport/WindowsLegacyConsoleOutput.cs
```diff
@@ -29,27 +26,6 @@ public override void PrepareConsole()
             _consoleBuffer.SetAsActiveBuffer();
         }
 
-        public override void Print(PixelBufferCoordinate bufferPoint, Color background, Color foreground,
-            FontStyle? style, FontWeight? weight, TextDecorationLocation? textDecoration, string str)
-        {
-            base.Print(bufferPoint, background, foreground, style, weight, textDecoration, str);
-
-            if (textDecoration == TextDecorationLocation.Underline)
```
jinek, line 37: This does not work, right?

tomlm: yeah

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
tomlm: Yep, renamed to WritePixel() so that it matches WriteText, much cleaner, it's an in paramter so no copy of data

src/Consolonia.Core/Infrastructure/DefaultNetConsoleOutput.cs
tomlm: removed

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
tomlm: Because this is called before avalonia is started up.  I want to clear the screen of the test chars while waiting for avalonia to spool up and do the first render.

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
tomlm: It doesn't matter WriteText(x) => sb.Append(x)

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -12,14 +12,24 @@ namespace Consolonia.Core.Infrastructure
     /// <summary>
     ///     Console implementation which uses ANSI escape sequences for output
     /// </summary>
+    /// <remarks>
+    ///     This console buffers all output and only writes to the console on Flush.
```
jinek, line 16: This is not true. For instance `WriteText` does not buffer

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -12,14 +12,24 @@ namespace Consolonia.Core.Infrastructure
...
     {
         private const string TestEmoji = "👨‍👩‍👧‍👦";
 
         private static readonly Lazy<IConsoleColorMode> ConsoleColorMode =
             new(() => AvaloniaLocator.Current.GetRequiredService<IConsoleColorMode>());
 
+        private readonly StringBuilder _outputBuffer = new();
+
         private PixelBufferCoordinate _headBufferPoint;
+        private Color _lastBackground = Colors.Transparent;
```
jinek, line 28: These you could also store as a Pixel. Close this comment yourself please

jinek (approved):

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
jinek: Close this yourself please

tomlm: WriteText does buffer, I changed it to.

PR #540, opened 2025-11-21
add flush() to command operations.

tomlm:
commands can be outside of render loop and didn't flush()


src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -267,18 +268,21 @@ public void SetCaretStyle(CaretStyle caretStyle)
         public void HideCaret()
```
jinek, line 267: We should think how to name it to understand what is immediate and what's buffered

tomlm: yeah

PR #548, opened 2025-11-23
Removed Variation sniffing for icons

tomlm:
Icons should now work reliably. When drawing complex characters we explicitly draw blank for the width, then the emoji, then move the caret to where want it to be. With buffering this costs nothing and it makes it very robust. 

I was able to remove variation sniffing, flags.

I also made the clipboard fall back work, on my system I have DISPLAY but no X11, I think this is more robust way to do it.

Fixes: #549 , #550


src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -151,33 +138,16 @@ public void WritePixel(PixelBufferCoordinate position, in Pixel pixel)
...
-                        }
-
-                        WriteText(Esc.SetCursorPosition(position.X, position.Y));
-                        WriteText(grapheme.Glyph);
-
-                        position =
-                            new PixelBufferCoordinate((ushort)(position.X + glyphWidth), position.Y);
-                    }
-                }
+                // We write out blank chars because we don't know how many cells will be rendered by the terminal
```
jinek, line 136: Why do you assume it will not be more than pixel.Width?

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
```diff
@@ -151,33 +138,16 @@ public void WritePixel(PixelBufferCoordinate position, in Pixel pixel)
...
+                // then we draw the complex glyph on top of the blank chars.
+                WriteText(new string(' ', pixel.Width));
+                SetCaretPosition(position);
+                WriteText(pixel.Foreground.Symbol.Complex);
```
jinek, line 143: A understand and appreciate this approach. Just as an idea: would not be better to keep a dictionary of probes (similar to what you did in Symbol)? Once we see no probe - we flush that complex glyph and same time reading the position - this way any next time we could already re-use the real width and do 0 additional instructions.

jinek:
I understand the idea, just to mention some issues related to complex glyph still remains:

1) it overwrites following pixels
image

2) Probably because of same reason they don't get refreshed after:
image

3) in textbox it's somehow messed, also happens when just moving the caret:
image

4) In DefaultNetConsole issue is still remaining:
image
image
image

5) Also issue remains with AnsiConsole under Linux Console:
image

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
```diff
@@ -78,18 +78,22 @@ public static AppBuilder UseAutoDetectClipboard(this AppBuilder builder)
...
                 if (IsWslPlatform())
                     clipboardImpl = new WslClipboard();
-                else if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DISPLAY")))
-                    clipboardImpl = new X11Clipboard();
                 else
                     try
                     {
-                        // alternatively use xclip CLI tool
-                        clipboardImpl = new XClipClipboard();
+                        clipboardImpl = new X11Clipboard();
```
jinek: I think if there is no DISPLAY variable we should not try to load X11 library.

src/Consolonia.Core/Infrastructure/AnsiConsoleOutput.cs
tomlm: I actually tried this, once we switch to cursor mode I can't use Console API, I don't know how to get the current position once in ansi mode.

jinek: Can we just use `Console.GetCursorPosition()` ?

PR #571, opened 2025-12-03
Add TTY mouse support via GpmConsole

tomlm:
Added **GpmMonitor**() which consumes mouse events from GPM
Changed CursesConsole to look at $TERM to decide to use GPM, as well as old ncurses.
This also fixes #569

### DIFF
--- a/src/Consolonia.PlatformSupport/GpmMonitor.cs
+++ b/src/Consolonia.PlatformSupport/GpmMonitor.cs
@@ -50,7 +50,7 @@ namespace Consolonia.PlatformSupport
         private int _gpmFd;
 
         /// <summary>
-        /// Initialize GPM Monitor, throws InvalidOperationException on failure to connect.
+        ///     Initialize GPM Monitor, throws InvalidOperationException on failure to connect.
         /// </summary>
         /// <exception cref="InvalidOperationException"></exception>
         public GpmMonitor()
@@ -73,7 +73,7 @@ namespace Consolonia.PlatformSupport
             _pumpTask = PumpGpmEventsAsync(_gpmCancellation.Token);
         }
 
-        /// <inheritdoc/>
+        /// <inheritdoc />
         public void Dispose()
         {
             Dispose(true);
@@ -81,13 +81,13 @@ namespace Consolonia.PlatformSupport
         }
 
         /// <summary>
-        /// Raised when a GPM mouse event is received (invoked on the UI thread).
+        ///     Raised when a GPM mouse event is received (invoked on the UI thread).
         /// </summary>
         /// <remarks>
-        /// * eventType - Pointer event type.
-        /// * point - Pointer position in console coordinates (0-based).
-        /// * wheelDelta - Wheel delta 
-        /// * modifiers - Keyboard/buttons modifier state
+        ///     * eventType - Pointer event type.
+        ///     * point - Pointer position in console coordinates (0-based).
+        ///     * wheelDelta - Wheel delta
+        ///     * modifiers - Keyboard/buttons modifier state
         /// </remarks>
         public event Action<RawPointerEventType, Point, Vector?, RawInputModifiers> MouseEvent;
 


src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -203,11 +209,23 @@ private void RenderToDevice()
             }
         }
 
-        private static Color GetInvertColor(Color color)
+
+        private static Color GetContrastColor(Color color)
```
jinek, line 229: ❤️

src/Consolonia.Gallery/App.cs
```diff
@@ -23,9 +23,18 @@ public override void OnFrameworkInitializationCompleted()
             if (((ConsoloniaLifetime)ApplicationLifetime).IsRgbColorMode()
                 && !((ConsoloniaLifetime)ApplicationLifetime).Args.Any(argument => argument != null &&
                     argument.EndsWith(TurboVisionProgramParameter, StringComparison.OrdinalIgnoreCase)))
+            {
                 Styles.Add(new ModernTheme());
+            }
             else
-                Styles.Add(new TurboVisionTheme());
+            {
+                bool isTTY = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DISPLAY")) &&
```
jinek: Misleading variable name. Should be "IsNotLinuxDesktop" or better "norX11NeitherWayland"

src/Consolonia.Gallery/App.cs
```diff
@@ -23,9 +23,18 @@ public override void OnFrameworkInitializationCompleted()
...
+                             string.IsNullOrEmpty(Environment.GetEnvironmentVariable("WAYLAND_DISPLAY"));
+                if (isTTY)
```
jinek: Why do you want different themes?

src/Consolonia.PlatformSupport/CursesConsole.cs
```diff
@@ -207,27 +214,31 @@ public override void PrepareConsole()
...
 
-            _supportsMouse = mouseMask != 0;
-            _supportsMouseMove = mouseMask.HasFlag(Curses.Event.ReportMousePosition);
+                _supportsMouse = mouseMask != 0;
+                _supportsMouseMove = mouseMask.HasFlag(Curses.Event.ReportMousePosition);
 
-            Curses.mouseinterval(0); // if we don't do this mouse events are dropped
-            Curses.timeout(NoInputTimeout);
+                Curses.mouseinterval(0); // if we don't do this mouse events are dropped
+                Curses.timeout(NoInputTimeout);
```
jinek: It's already set further

src/Consolonia.PlatformSupport/CursesConsole.cs
```diff
@@ -608,107 +612,70 @@ private void HandleMouseInput(Curses.MouseEvent ev)
...
-                        RaiseMouseClickedEvent(RawPointerEventType.MiddleButtonDown,
-                            RawPointerEventType.MiddleButtonUp, 1, point, rawInputModifiers);
-                        break;
-                    case Curses.Event.Button2DoubleClicked:
-                        RaiseMouseClickedEvent(RawPointerEventType.MiddleButtonDown,
-                            RawPointerEventType.MiddleButtonUp, 2, point, rawInputModifiers);
-                        break;
-                    case Curses.Event.Button2TripleClicked:
-                        RaiseMouseClickedEvent(RawPointerEventType.MiddleButtonDown,
-                            RawPointerEventType.MiddleButtonUp, 3, point, rawInputModifiers);
```
jinek, line 633: Should we exclude this from mousemask during initialization?

src/Consolonia.PlatformSupport/CursesConsole.cs
```diff
@@ -608,107 +612,70 @@ private void HandleMouseInput(Curses.MouseEvent ev)
...
-                    case Curses.Event.Button3TripleClicked:
-                    case Curses.Event.Button4TripleClicked:
-                        RaiseMouseClickedEvent(RawPointerEventType.RightButtonDown,
-                            RawPointerEventType.RightButtonUp, 3, point, rawInputModifiers);
-                        break;
                     case Curses.Event.Button1Pressed:
-                        _moveModifers = rawInputModifiers | RawInputModifiers.LeftMouseButton;
-                        RaiseMouseEvent(RawPointerEventType.LeftButtonDown, point, wheelDelta, _moveModifers);
+                        RaiseMouseEvent(RawPointerEventType.LeftButtonDown, point, null,
+                            rawInputModifiers | RawInputModifiers.LeftMouseButton);
```
jinek: I asked you this question also. Why do we pass the state one more time into modifiers, should not be they empty i no CTRL or SHIFT etc were pressed?

src/Consolonia.PlatformSupport/CursesConsole.cs
```diff
@@ -608,107 +612,70 @@ private void HandleMouseInput(Curses.MouseEvent ev)
...
-                            RawInputModifiers.None);
+                        RaiseMouseEvent(RawPointerEventType.XButton1Up, point, null,
+                            rawInputModifiers | RawInputModifiers.XButton1MouseButton);
                         break;
-
                     case Curses.Event.ReportMousePosition:
-                        RaiseMouseEvent(RawPointerEventType.Move, point, wheelDelta,
-                            rawInputModifiers | _moveModifers);
-                        break;
+                        if (point != _lastPoint)
```
jinek: Why should we skip this?

src/Consolonia.PlatformSupport/Gpm.cs
```diff
@@ -0,0 +1,255 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Consolonia.PlatformSupport
+{
+    /// <summary>
+    ///     GPM event types
```
jinek, line 9: 😂

src/Consolonia.Gallery/App.cs
tomlm: reverted

src/Consolonia.Gallery/App.cs
tomlm: reverted

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: fixed

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: On my TTY I was getting mouse move events on the pixel level, which causes us to push a ton of events. For some reason the TTY mouse is already laggy (I would love input on how to make better) so I'm just trying to skip things that don't cause anything to happen.

src/Consolonia.PlatformSupport/Gpm.cs
```diff
@@ -0,0 +1,255 @@
...
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct GpmConnect
+    {
+        public ushort EventMask; // Event types to receive
+        public ushort DefaultMask; // Default event mask
+        public ushort MinMod; // Minimum modifier
+        public ushort MaxMod; // Maximum modifier
+        public int Pid; // Process ID
+        public int VirtualConsole; // Virtual console
```
jinek: I'm proposing to move this comments out to XML comments in this file.

src/Consolonia.PlatformSupport/Gpm.cs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+
+    [StructLayout(LayoutKind.Sequential)]
+    public struct Timeval
+    {
+        public long Sec; // Changed from int to long for 64-bit compatibility
```
jinek: I don't understand these comments

src/Consolonia.PlatformSupport/Gpm.cs
```diff
@@ -0,0 +1,255 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Consolonia.PlatformSupport
+{
+    /// <summary>
+    ///     GPM event types
+    /// </summary>
+    [Flags]
+    public enum GpmEventType : ushort
```
jinek: I bet these structures can be internal?

src/Consolonia.PlatformSupport/Gpm.cs
```diff
@@ -0,0 +1,255 @@
...
+        /// </summary>
+        /// <param name="x">X position (-1 to hide cursor)</param>
+        /// <param name="y">Y position</param>
+        /// <param name="flag">Drawing flag</param>
+        /// <returns>0 on success</returns>
+        [DllImport(LibraryName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "Gpm_DrawPointer")]
+        public static extern int DrawPointer(int x, int y, int flag);
+
+        [DllImport("libc", EntryPoint = "select", SetLastError = true)]
+        public static extern int Select(int nfds, IntPtr readfds, IntPtr writefds, IntPtr exceptfds,
```
jinek: dotnet already has Socket.Select - isn't that same?

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: I don't think I understand your question. 
The foreach() at the top enumerates every flag that is checked (which seems pretty inefficient to me) The order of that is not deterministic, so they compute the keyboard state before the loop, then ignore them in the loop and then pass it on to each event.  Seems a bit complicated but that was the curses code we inherited.
I think we could get rid of the loop entirely with the mousemask to UP|DOWN|MOVE|WHEEL|MODIFIERS, I never see them get combined.

src/Consolonia.PlatformSupport/GpmConsole.cs
```diff
@@ -0,0 +1,301 @@
...
+            if (gpmEvent.Modifiers.HasFlag(GpmModifiers.Alt))
+                modifiers |= RawInputModifiers.Alt;
+
+            // Add button state to modifiers for regular mouse events
+            if (gpmEvent.Buttons.HasFlag(GpmButtons.Left))
+                modifiers |= RawInputModifiers.LeftMouseButton;
+            if (gpmEvent.Buttons.HasFlag(GpmButtons.Middle))
+                modifiers |= RawInputModifiers.MiddleMouseButton;
+            if (gpmEvent.Buttons.HasFlag(GpmButtons.Right))
+                modifiers |= RawInputModifiers.RightMouseButton;
```
jinek: There is FlagTranslator - you could simplify this calculation of modifiers with that

src/Consolonia.PlatformSupport/GpmConsole.cs
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    if (_gpmFd >= 0)
+                    {
+                        _ = Gpm.Close();
+                        _gpmFd = -1;
+                    }
+                }
+                catch (DllNotFoundException)
+                {
+                    // Library unloaded, ignore
```
jinek: Just curios how can this happen when the process is running?

src/Consolonia.PlatformSupport/GpmConsole.cs
```diff
@@ -0,0 +1,301 @@
...
+            catch (EntryPointNotFoundException)
+            {
+            }
+        }
+
+        private async Task GpmEventLoop()
+        {
+            await Helper.WaitDispatcherInitialized();
+
+            while (!_gpmCancellation.Token.IsCancellationRequested && !Disposed)
```
jinek: I'm almost sure left part of condition can throw ObjectDisposedException

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
```diff
@@ -192,5 +192,22 @@ private static bool IsWindowsTerminal()
...
         }
+
+        /// <summary>
+        ///     Creates the appropriate console implementation for Unix/MacOSX platforms.
+        ///     Prefers GpmConsole if GPM is available in TTY environments, otherwise falls back to CursesConsole.
+        /// </summary>
+        private static IConsole CreateUnixConsole()
+        {
+            // Check if we're in a TTY environment (not X11/Wayland)
+            // Resharper disable InconsistentNaming
```
jinek: Add TTY to dictionary since we using it in many places

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -135,9 +135,15 @@ private void RenderToDevice()
...
                         _consoleCursor.Coordinate.Y == y &&
                         _consoleCursor.Coordinate.X <= x && x < _consoleCursor.Coordinate.X + _consoleCursor.Width)
-                        pixel = pixel.Blend(new Pixel(new PixelForeground(
-                            new Symbol(_consoleCursor.Type[x - _consoleCursor.Coordinate.X], 1),
-                            GetInvertColor(pixel.Background.Color))));
+                    {
+                        char cursorChar = _consoleCursor.Type[x - _consoleCursor.Coordinate.X];
+                        // only if we are drawing a " " and the pixel underneath is not wide char
+                        // do we lift the character from the underlying pixel and invert it
+                        if (_consoleCursor.Type == " " && pixel.Width == 1)
```
jinek, line 139: Why do you want space instead of empty string?

jinek: ah, I understood. But we should not draw arrow cursor when normal OS cursor is supported

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
```diff
@@ -192,5 +192,22 @@ private static bool IsWindowsTerminal()
...
+            bool isTTY = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DISPLAY")) &&
+                         string.IsNullOrEmpty(Environment.GetEnvironmentVariable("WAYLAND_DISPLAY"));
+            // If in TTY and GPM is available, use GpmConsole for better mouse support
+            if (isTTY && Gpm.IsGpmAvailable()) return new GpmConsole();
```
jinek: With this logic, if GPM installed and I connect via SSH - it tries to ues Gpm, in my case clicks stop working. I guess we should examine what curses HasMouse and mask return instead of relying on these variables. Their absence don't mean mouse is not working in curses.

src/Consolonia.PlatformSupport/CursesConsole.cs
jinek: Yes. at least ncurses specify that only 1 bit is set in each event.

No, my question is about why do we set modifiers to LEFTMOUSEBUTTON when we send LeftButtonDown event?

src/Consolonia.PlatformSupport/CursesConsole.cs
jinek: See, I'm considering following scenario: I switched windows or opened a dropdown/menu for instance. Then I do small mouse move (within same pixel), should UI process that move and highlight an element under mouse?

jinek:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> 
> @jinek: The walkthrough appears because several of its components are still enabled in your configuration. To completely disable it, add these settings to your `.coderabbit.yaml`:
> 
> ```yaml
> reviews:
>   assess_linked_issues: false
>   related_issues: false
>   related_prs: false
>   suggested_labels: false
> ```
> 
> Your configuration already has:
> 
>     * `high_level_summary: false` ✓
> 
>     * `changed_files_summary: false` ✓
> 
>     * `sequence_diagrams: false` ✓
> 
>     * `estimate_code_review_effort: false` ✓
> 
> 
> The remaining enabled components (issue assessment, related PRs, and suggested labels) are what's causing the walkthrough to still appear. Disabling them will prevent the walkthrough comment from being posted.

But we would like to keep that. Can you just memorize to keep suggested labels and relevant prs etc, but exclude walkthgouth itself and checks and "thank you" ?

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: I have rewritten this code again. I had a regression because I wasn't tracking _moveModifiers. For drag to work we need to remember modifier flags and merge them into the MOVE event.

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: yes, I updated the mousemask, good call

src/Consolonia.PlatformSupport/Gpm.cs
tomlm: yep, making them internal

src/Consolonia.PlatformSupport/Gpm.cs
tomlm: Um. Perhaps I used an AI to generate them? :) 
Removed

src/Consolonia.PlatformSupport/Gpm.cs
tomlm: removed them, they are redundent.

src/Consolonia.PlatformSupport/Gpm.cs
tomlm: I didn't know that, but yes it is, I have removed this definition and changed the code to use Socket.Select

src/Consolonia.PlatformSupport/GpmConsole.cs
tomlm: <sigh> Yes I didn't know about that, I will look at this.

tomlm: now using flagtranslator

src/Consolonia.PlatformSupport/GpmConsole.cs
```diff
@@ -0,0 +1,301 @@
...
+
+                // Pattern 2: buttons=0, type=MOVE = scroll DOWN
+                if (gpmEvent.Type == GpmEventType.Move &&
+                    gpmEvent.Buttons == GpmButtons.None)
+                {
+                    // Debug.WriteLine("GPM: Wheel DOWN detected (MOVE+no buttons pattern)");
+                    RaiseMouseEvent(RawPointerEventType.Wheel, point, new Vector(0, -1), modifiers);
+                    return;
+                }
+            }
```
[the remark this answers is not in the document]
tomlm: this is reflecting the state of what GPM produces and so is correct.

src/Consolonia.PlatformSupport/GpmConsole.cs
tomlm: AI generated this. It does seem a bit silly.

tomlm: I have removed them

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
tomlm: Where do you think would be a good place for IsTTY to live?
Curses will think it has the mouse when GPM is running. 
Curses mostly works with GPM except 
* Click drag does selection instead of generating events, so you can't select text, resize windows etc.
* Wheel Scroll is completely not working

src/Consolonia.PlatformSupport/GpmConsole.cs
tomlm: are you saying I need to wrap in try/catch?

src/Consolonia.PlatformSupport/GpmConsole.cs
```diff
@@ -0,0 +1,277 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.Sockets;
```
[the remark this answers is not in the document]
tomlm: good catch on the using

src/Consolonia.PlatformSupport/Gpm.cs
tomlm: also made it a member and dispose of it

src/Consolonia.PlatformSupport/GpmConsole.cs
tomlm: updated

src/Consolonia.PlatformSupport/PlatformSupportExtensions.cs
tomlm: I couldn't find any other place using it.

src/Consolonia.PlatformSupport/CursesConsole.cs
```diff
@@ -207,27 +259,60 @@ public override void PrepareConsole()
...
+                    _supportsMouseMove = true;
+                }
+                catch (DllNotFoundException)
+                {
+                    // ignore
+                }
+                catch (EntryPointNotFoundException)
+                {
+                    // ignore
+                }
```
tomlm: fixed

src/Consolonia.Controls/ControlUtils.cs
```diff
@@ -35,7 +35,8 @@ public static ushort MeasureText(this string text)
             ArgumentNullException.ThrowIfNull(text);
 
             IConsoleCapabilities console = ConsoleCapabilities.Value;
-            bool supportsComplexEmoji = console == null || console.SupportsComplexEmoji;
+            bool supportsComplexEmoji = console == null ||
```
jinek: Why null means it supports complex emoji?

src/Consolonia.Controls/IConsoleCapabilities.cs
```diff
@@ -1,10 +1,43 @@
...
+
+        /// <summary>
+        ///     Supports detection of Alt key by itself
+        /// </summary>
+        SupportsAltSolo = 0x01,
+
+        /// <summary>
+        ///     Console supports general mouse input
+        /// </summary>
+        SupportsMouse = 0x02,
```
jinek: No, but this is wrong. I'm proposing not to use flags, but just enums. This one is not better than boolean values. These flags also allow 8 states of mouse while we need only 4.

jinek: After a discussion decided that we are fine to have mouseCursor with no mouseMoves

jinek: Just discussed this idea:

MouseClicks = 1,
MouseMove = 3
MouseCursor = 7


But may be binary writing would be better

src/Consolonia.Controls/IConsoleCapabilities.cs
```diff
@@ -1,10 +1,43 @@
...
+        /// </summary>
+        SupportsMouseCursor = 0x08,
+
+        /// <summary>
+        ///     Supports complex composite emoji rendering
+        /// </summary>
+        SupportsComplexEmoji = 0x10
+    }
+
     public interface IConsoleCapabilities
```
jinek, line 38: Why do we need this interface. It always belongs to IConsole only, nothing else possible, right? We could just have a property in IConsole

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -135,9 +135,31 @@ private void RenderToDevice()
                     if (!_consoleCursor.IsEmpty() &&
                         _consoleCursor.Coordinate.Y == y &&
                         _consoleCursor.Coordinate.X <= x && x < _consoleCursor.Coordinate.X + _consoleCursor.Width)
-                        pixel = pixel.Blend(new Pixel(new PixelForeground(
-                            new Symbol(_consoleCursor.Type[x - _consoleCursor.Coordinate.X], 1),
-                            GetInvertColor(pixel.Background.Color))));
+                    {
+                        if (_consoleCursor.Type == " " && pixel.Width == 1)
```
jinek, line 139: Second part of condition is always true if first is true

src/Consolonia.Core/Infrastructure/ConsoleBase.cs
```diff
@@ -128,9 +125,8 @@ public PixelBufferSize Size
             }
         }
 
-        public virtual bool SupportsComplexEmoji => _consoleOutput.SupportsComplexEmoji;
 
-        public abstract bool SupportsAltSolo { get; }
+        public ConsoleCapabilities Capabilities { get; protected set; }
```
jinek, line 128: Before we were requesting each time. Now setting in the ctor. Do we know it's initialized there once? PrepareConsole or other stuff does not change the value?

src/Consolonia.Controls/IConsoleCapabilities.cs
tomlm: See update

src/Consolonia.Controls/ControlUtils.cs
tomlm: oh. 

Crap. That's a skeleton unit test hack.

I have a bunch of unit tests (like AsciArt, etc.) which only have dependency on MeasureText.  MeasureText has a dependency IConsole which is in AvaloniaLocator.  I don't have avalonia locator hooked up at all for these tests.

I'll update unit tests to register a proper mock for this.

src/Consolonia.PlatformSupport/GpmMonitor.cs
```diff
@@ -0,0 +1,154 @@
...
+                throw new InvalidOperationException("Failed to open GPM connection");
+
+            _gpmInitialized = true;
+            _pumpTask = PumpGpmEventsAsync(_gpmToken);
+        }
+
+        public void Dispose()
+        {
+#pragma warning disable CA1063 // Implement IDisposable Correctly
+#pragma warning disable CA1303 // Do not pass literals as localized parameters
```
jinek: no literals here

src/Consolonia.Controls/IConsoleCapabilities.cs
jinek: This is out of scope

src/Consolonia.Core/Drawing/RenderTarget.cs
tomlm: No, the second part is when we are trying to put a " " over the top of a wide char. The pixel is the symbol underneath the cursor

src/Consolonia.Core/Infrastructure/ConsoleBase.cs
tomlm: PrepareConsole is where it gets set because we do our capabilities sniffing in the prepareconsole

src/Consolonia.PlatformSupport/CursesConsole.cs
tomlm: I don't skip this anymore, so I'm resolving.

src/Consolonia.PlatformSupport/GpmMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+using Avalonia.Threading;
+using Consolonia.Core.Helpers;
+using Consolonia.Core.InternalHelpers;
+
+namespace Consolonia.PlatformSupport
+{
+    /// <summary>
+    ///     implementation with GPM (General Purpose Mouse) support for TTY environments
+    /// </summary>
+    internal class GpmMonitor : IDisposable
```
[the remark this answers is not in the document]
tomlm: updated

PR #578, opened 2025-12-26
Comment out the non-functional hyperlink section in GalleryButton.axaml, add a TODO for replacement.

jinek:


src/Consolonia.Gallery/Gallery/GalleryViews/GalleryButton.axaml
```diff
@@ -68,6 +68,8 @@
             <TextBlock TextWrapping="Wrap">A button control that functions as a navigateable hyperlink.</TextBlock>
 
 
+            <!--
+            this does not work with no DISPLAY variable (todo: implement replace of BclLauncher)
```
tomlm, line 72: This does work normally though, right? Do we really need to comment out?

tomlm (approved):

src/Consolonia.Gallery/Gallery/GalleryViews/GalleryButton.axaml
jinek: It did not work somehow. May be it does not work in TTY. Lets keep this PR for some time. I need to recall what happened

jinek: So yes.  Links don't work in tty which is expected.  But also there is no focus indicator on the links.

PR #581, opened 2026-01-03
Optimize Mouse movement

tomlm:
TTY is much slower than GUI terminal and mouse movement is really slow. 

I spent a bunch of time working the system, and it came down to the way we are attempting to render the cursor. 

We monitor cursor movement and every time the mouse moves, we render the pixelbuffer.

I changed this logic to only render every 16ms, giving us 60fps animation for mouse movement (and BTW for dragging windows too).  

I also found that we are doing sequential linq query on every color lookup so I added a cache.


src/Consolonia.Core/Drawing/PixelBufferImplementation/EgaConsoleColor/EgaConsoleColorMode.cs
```diff
@@ -114,12 +116,16 @@ public Color Blend(Color color1, Color color2, bool isTargetForeground)
 
         private ConsoleColor MapToConsoleColor(Color color, bool isForeground)
         {
+            if (_consoleColorCache.TryGetValue(color, out ConsoleColor cachedColor))
+                return cachedColor;
             int r = color.R, g = color.G, b = color.B;
 
             // Find the nearest ConsoleColor by RGB distance
-            return GetPalette(isForeground)
+            ConsoleColor consoleColor = GetPalette(isForeground)
```
jinek, line 120: Does this have any impact currently, because pallete is just 8 or 16 colors, while cache grows with every posisble color which is 10 bytes. it can grow maximum 200 mb. I'm proposing to make more analysis here to check how it impact performance and memory.

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -265,7 +273,17 @@ private void OnCursorChanged(ConsoleCursor consoleCursor)
             _consoleTopLevelImpl.DirtyRegions.AddRect(oldCursorRect);
             _consoleTopLevelImpl.DirtyRegions.AddRect(newCursorRect);
 
-            RenderToDevice();
+            if (!_renderPending)
+            {
+                _renderPending = true;
+
+                // this gates rendering of cursor to (60fps) to avoid excessive rendering when moving cursor fast
+                DispatcherTimer.RunOnce(() =>
```
jinek, line 304: I wanted to propose to use some RateLimiter to avoid lag. But this is better than another NUGET package. BTW we are having 120 FPS, I don't know why https://github.com/Consolonia/Consolonia/blob/6f856e052f2bcd5b6f63d3dc6e0fe80d06c2d887/src/Consolonia.Core/Infrastructure/ConsoloniaPlatform.cs#L53

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -265,7 +273,17 @@ private void OnCursorChanged(ConsoleCursor consoleCursor)
...
+                {
+                    _renderPending = false;
+                    RenderToDevice();
```
jinek: May be it's better to jump back to ThreadPool to unblock the dispatcher? Or am I wrong here? It was Dispatcher thread even before?

src/Consolonia.Core/Drawing/PixelBufferImplementation/EgaConsoleColor/EgaConsoleColorMode.cs
tomlm: We call this for every single pixel on every render. I was not running the profiler (I don't know how do to this in TTY), but just timing the render loop, and the render loop went from ~20ms to ~5ms.  But I did not do a formal profile. 

I can do a separate PR for this if you want.

src/Consolonia.Core/Drawing/RenderTarget.cs
tomlm: DispatcherTimer seems to exist for exactly this sort of scenario.

src/Consolonia.Core/Drawing/PixelBufferImplementation/EgaConsoleColor/EgaConsoleColorMode.cs
```diff
@@ -170,9 +193,19 @@ public Color ConvertToAvaloniaColor(ConsoleColor consoleColor, bool isForeground
...
-                    (ConsoleColor _, (int R, int G, int B) rgb) =
-                        GetPalette(isForeground).First(c => c.Color == consoleColor);
-                    return Color.FromRgb((byte)rgb.R, (byte)rgb.G, (byte)rgb.B);
+                {
+                    (ConsoleColor Color, (int R, int G, int B) Rgb)[] palette = GetPalette(isForeground);
+
+                    for (int i = 0; i < palette.Length; i++)
+                    {
+                        (ConsoleColor Color, (int R, int G, int B) Rgb) entry = palette[i];
+                        if (entry.Color == consoleColor)
```
jinek: "First()" does not do any allocations

src/Consolonia.Core/Drawing/RenderTarget.cs
```diff
@@ -265,7 +266,17 @@ private void OnCursorChanged(ConsoleCursor consoleCursor)
             _consoleTopLevelImpl.DirtyRegions.AddRect(oldCursorRect);
             _consoleTopLevelImpl.DirtyRegions.AddRect(newCursorRect);
 
-            RenderToDevice();
+            if (!_renderPending)
```
jinek, line 299: With this logic cursor is skipped at last move sometimes. On video after each move I stop the mouse and hold it a few seconds before clicking the button so cursor jumps to real position 

https://github.com/user-attachments/assets/68910dff-ed5a-48cc-a008-7d90b67f203b

jinek: ~I've tried to find something for this task in Avalonia and closest is only `IRenderTimer`. As an option you can submit to IRenderTimer.Tick and if a flag `_requestRedraw` was true - call RenderToDevice.~

ah, this one also won't work this easy. Coz if flag was reset right after it was set - last redraw will get lost again.

tomlm: I can't reproduce this. Is it possible this is the sandbox you are running in?
What happens in TTY mode?

jinek: Try to increase the time interval to make it obvious.
That is I ran in just working desktop.
I think in TTY will be same.
Last move of the mouse happens when `_renderPending` is true which means it's just skipped.

src/Consolonia.Core/Drawing/PixelBufferImplementation/EgaConsoleColor/EgaConsoleColorMode.cs
tomlm: if it's orderby.FirsT() then it has to sort the entire list before it can return the first element.

src/Consolonia.Core/Drawing/RenderTarget.cs
tomlm: fuck. It's not my change. On linux curses is delivering mouse events out of order.

tomlm: When I turn on the text cursor in win32 it is not reproducable at all. When I turn it on in terminal I can reproduce it. If you look at the event logs you will see that an old event gets played last and so the cursor is correct for the sequence but the sequence is wrong.

src/Consolonia.Core/Drawing/PixelBufferImplementation/EgaConsoleColor/EgaConsoleColorMode.cs
jinek: right. then MinBy/Min don't do full sort

src/Consolonia.Core/Drawing/RenderTarget.cs
jinek: thats bad/

jinek (approved):
Just a though. I don't like that we are adding lag here. With first move there is no need for a delay. Also I'm feeling that its somehow possible to trigger Avalonia to call re-render according to internal timer which is set to own FPS.

Issue #602, opened 2026-03-28
CJK characters rendered incorrectly due to Emoji Variation Selector (U+FE0F) appended to wide characters


BruceChenQAQ:
### Observed behavior and comparison results

1. Running by double-clicking
<img width="634" height="177" alt="Image" src="https://github.com/user-attachments/assets/157d3214-96a1-4699-848c-e47d2df1f831" />

<img width="572" height="244" alt="Image" src="https://github.com/user-attachments/assets/65aefd87-fd37-415e-859d-e2fa5eb3429e" />

> Copied straight from the terminal—when I type, the text just turns blank, and copying it gives blank text instead of the actual characters.
> ```
>  ♡  ♡  ♡  ♡  ♡  ♡  ♡  ♡  ♡  ♡  0.0                             ○  17
> Minecraft Console Client v26.1 - for MC 1.4.6 to 26.1 - Github.com/MCCTeam
> 帮️助️我️们️翻️译️MCC  https://crwd.in/minecraft-console-client
> 您️正️在️使️用️离️线️模️式️
> 正️在️获️取️          ...
>       的️Minecraft      1.21.11 (    v774)
> [MCC] 该️
>       录️...
> [MCC] 服️务️器️正️处️于️离️线️模️
> [MCC]
>      '/quit'
> [MCC]              '/respawn'
> 
> ```

2. Running from CMD or PowerShell after launching it first

<img width="605" height="241" alt="Image" src="https://github.com/user-attachments/assets/655d66b4-6ecf-43a4-b96a-5fe066b3037d" />

> ```
> Minecraft Console Client v26.1 - for MC 1.4.6 to 26.1 - Github.com/MCCTeam
> 帮助我们翻译MCC：https://crwd.in/minecraft-console-client
> 您正在使用离线模式。
> 正在获取服务器信息...
> 服务器的Minecraft版本：1.21.11 (协议v774)
> [MCC] 该版本受到支持
> 正在登录...
> [MCC] 服务器正处于离线模式。
> [MCC] 已成功加入服务器。
> 输入 '/quit' 离开服务器。
> [MCC] 你死了！输入 '/respawn' 重生。
> ```

3. Running by double-clicking without using Consolonia

<img width="596" height="232" alt="Image" src="https://github.com/user-attachments/assets/48a524e1-dbed-4043-b8c8-eae2c185964e" />

### Description

[...]

PR #624, opened 2026-04-12
[codex] Fix wide glyph output in legacy Windows console

BruceChenQAQ:
Related: #602

## Summary
- narrow emoji variation selector injection so only real emoji and wide symbol glyphs get `U+FE0F`
- skip zero-width placeholder cells in `DefaultNetConsoleOutput` so the Windows legacy console path does not write `NUL` bytes for wide glyph tails
- add regression coverage for full-width punctuation and legacy-console wide-glyph output

## Root Cause
`Symbol(char)` previously treated every width-2 non-letter as an emoji-style glyph and appended `U+FE0F`. That was too broad: full-width punctuation such as `：` `。` `（` `）` `！` is naturally double-width in CJK terminals and must stay as plain text.

A second issue remained in the legacy Windows console path. Wide glyph rendering uses zero-width placeholder cells for the trailing columns of a double-width glyph, but `DefaultNetConsoleOutput.WritePixel()` still wrote those placeholder cells. On the .NET console API this serialized a `\0` character into the output stream, which corrupted subsequent text layout in `WindowsLegacyConsoleOutput`.

## Fix
- restrict emoji-style variation injection to actual emoji and `UnicodeCategory.OtherSymbol` wide glyphs
- return early for `pixel.Width <= 0` in `DefaultNetConsoleOutput.WritePixel()`, matching the ANSI writer's behavior
- add tests covering both the punctuation case and the zero-width placeholder case

## Validation
- `dotnet test src/Tests/Consolonia.Core.Tests/Consolonia.Core.Tests.csproj -c Release --filter "FullyQualifiedName~DefaultNetConsoleOutputTests|FullyQualifiedName~SimpleSymbolTests|FullyQualifiedName~GlyphTests"`
- `dotnet build src/Consolonia/Consolonia.csproj -c Release`
- downstream manual validation in Minecraft Console Client using a Windows self-contained single-file publish launched by double-click, which forces the legacy console output path


jinek:
@BruceChenQAQ Thank you. Could you please confirm whether you can click the failing build to see the error?

BruceChenQAQ:
@jinek Yes. I checked the failing build and the error was in `Consolonia.Gallery.Tests`: `FontsTests.DisplaysWideTermText` was still asserting the old incorrect WideTerm output with injected `U+FE0F` selectors.

The rendering fix itself was already correct. I updated the outdated test expectation in `5c531e6` and reran `dotnet test src/Consolonia.sln` locally:
- `Consolonia.Core.Tests`: 270 passed, 1 skipped
- `Consolonia.Gallery.Tests`: 40 passed, 2 skipped

That commit retriggered the PR checks.

src/Consolonia.Core/Infrastructure/DefaultNetConsoleOutput.cs
```diff
@@ -59,6 +59,9 @@ public virtual PixelBufferCoordinate GetCaretPosition()
 
         public virtual void WritePixel(PixelBufferCoordinate position, in Pixel pixel)
         {
+            if (pixel.Width <= 0)
```
jinek: What is the case of negative width symbol? Im confused here because we are skipping the rest of the code, which does flushing and other stuff.

jinek: @coderabbitai any comments here?

jinek: But further when writing to output, when flushing, isn't it expected to have 0 width cells?

BruceChenQAQ: Negative width can't really happen here. `pixel.Width` on this path is effectively only used as 0 / 1 / 2, so this guard is really about the `width == 0` case.

That zero-width case is the trailing placeholder cell of a wide glyph. On the legacy `Console` API path we want to skip it completely, otherwise we end up writing the placeholder char itself and that is what corrupts the layout.

Skipping it also does not miss any required flush, because there is nothing to emit for that cell, and the frame still ends with the normal final `_console.Flush()` in `RenderTarget`.

I pushed a small follow-up to make that clearer: changed the condition from `<= 0` to `== 0` and added a comment explaining the placeholder case.

jinek (approved):

jinek:
@coderabbitai why isnt this pr merged automatically?

jinek:
> 
> > [!TIP]
> > For best results, initiate chat on the files or code changes.
> 
> `@jinek` CodeRabbit does not merge pull requests — it is an AI code reviewer. Merging is a human action. To merge this PR, you (as a maintainer) can click **Merge pull request** on GitHub directly.
> 
> If you'd like me to approve it first, you can comment `@coderabbitai approve` on the PR.

Ive clickef automatic merge, but something prevents it.

@coderabbitai

PR #636, opened 2026-04-25
Sixsel support ALPHA (DO NOT MERGE)

tomlm:
sixel
tootoot

* implement sixsel
* store sixsels per cell, giving accurate redraw for any image cell
* rewrote AnsiConsoleOutput to use stdout stream <-- significant perf improvement for all rendering!
* changed image viewer to use high res images in carousel.


tomlm:
sixelframe
Just because I was curious

PR #646, opened 2026-05-05
Lamda linq statement was allocating megabytes for every render frame.

tomlm:
The baseline shows each Contains(string) call takes ~43–44 µs and allocates 128 bytes per call. The memory profiler confirms the root cause — System.Func<,> and SZGenericArrayEnumerator objects are being heap-allocated by the LINQ lambda inside Contains(string), totalling over ~62 MB and ~41 MB respectively across the benchmark run.
This matches what the profiler found earlier: the lambda rect => inclusive ? rect.Contains(point) : rect.ContainsExclusive(point) allocates:
•	A Func<PixelRect, bool> delegate (heap closure)
•	A <>c__DisplayClass (closure object capturing inclusive and Point)
•	A SZGenericArrayEnumerator for the .Any() call
The fix — replacing LINQ with a plain for loop — eliminates all three allocations entirely


src/.github/copilot-instructions.md
```diff
@@ -0,0 +1,12 @@
+# Copilot Instructions
+
+## General Guidelines
+- First general instruction
+- Second general instruction
```
jinek: 😂 Won't let it in

tomlm: lol, ok

jinek (approved):

PR #651, opened 2026-05-07
ai slop

jinek:
This reduce CPU consumption when idle from >2% to 0-0.5%
~Code is mostly AI slope. I've been trying to fix it which ends up with what you see. However, all this stuff is needed only for Avalonia 11 since Avalonia 12 does not have idle issue. We will remove this code~
[v] this did not work as pointed by code rabbit. Will go thru the code myself.


jinek:
Exit does not work with this implementation. Spending too much time for this, closing

PR #664, opened 2026-06-05
Gradient Fixes, Gradients on Lines/Boxes, LineBrush Animation

mstrobel:
### Gradient interpolation rewritten — `BrushExtensions.cs`

Replaced the broken `FromPosition` (which averaged separate horizontal/vertical positions and ignored `StartPoint`/`EndPoint`, `SpreadMethod`, and out-of-range stops) with `ColorAt(this IBrush, PixelPoint, PixelRect bounds)`:

- **Linear**: proper projection onto the StartPoint→EndPoint axis.
- **Radial**: focal-point ray/unit-circle solve (reduces correctly to distance-from-center when origin = center).
- **Conic**: real angle (`atan2`, clockwise from top, offset by `Angle`) — was entirely broken before.
- **SpreadMethod** Pad/Repeat/Reflect, applied over the gradient vector's [0,1] interval (not the stop range — this distinction matters for Repeat/Reflect when stops don't cover [0,1]).
- **Stops** needn't be sorted, may start/end anywhere, and pad to the first/last color outside their range.
- Unwraps `LineBrush` so a gradient wrapped in one is sampled; honors brush `Opacity`.

### Lines & rectangles paint with gradients — `DrawingContextImpl.Boxes.cs`

- `FillRectangleWithBrush` samples per-cell via `ColorAt`, mapping the gradient onto the shape's own bounds.
- Line drawing (`DrawBoxLine`/`DrawEdgeLine`/`DrawLineSymbolAndMoveHead`) now carries the brush and samples per-cell; `ExtractColorOrNull`… → `ExtractBrushOrNull`… returns the inner solid/gradient brush.
- Rectangle borders thread the **whole-rectangle bounds** into all four edges so they share one gradient; standalone lines fall back to the clip.

⚠️ _**Design note:** Gradients mapped to shape bounds, as opposed to `CurrentClip`; easily changed if desired._

### `LineBrush` may now be animated

- `LineBrushAnimator` interpolates the inner brush: gradients via Avalonia's own `GradientBrushAnimator` (so a wrapped gradient animates identically to an unwrapped one), solids via ARGB lerp; `LineStyle` switches at the midpoint.
- Registers itself into `BaseBrushAnimator`'s private `_brushAnimators` registry by reflection (replicating what the sealed `RegisterBrushAnimator` does), with the factory producing the animator via the `public ICustomAnimator.CreateWrapper()`.
- `LineBrush`'s static constructor calls `EnsureRegistered()` (thread-safe, never throws — failures are logged and degrade to discrete switching).

⚠️ _**Design note:** Since this relies on reflection, this set of changes is contained in its own commit. If you drop it, the gallery changes will still work; the animated border will fall back to using the direct `LinearGradientBrush` it had in the previous commit._

### Tests

- `BrushExtensionsTests` has 10 tests, including linear/horizontal-only, `Pad`/`Repeat`/`Reflect`, out-of-range stops, conic quadrants, radial, `LineBrush` unwrap).
- Added 3 new tests in `DrawingContextImplTests` (fill shape-bounds, standalone-line clip fallback, border edges spanning the rectangle). All values hand-computed.
- `LineBrushAnimatorTests` has 5 tests. Most important is `AvaloniaDrivesAnimatorForBorderBrushAnimation`, which drives a real `Border.BorderBrush` keyframe animation with a real Avalonia clock and asserts the brush is an interpolated `LineBrush` at 50% — proving Avalonia actually selects and ticks our animator, not just that the registry entry exists.

### Gallery

- Animated border added to Borders gallery (`LinearGradientBrush`, `Reflect`, offsets to `300%`); hits exactly the raw-gradient-pen + out-of-range-_t_ paths the new code handles.
- Added a few `Border` gradient examples to `GalleryGradientBrush.axaml`.
- Changed background of labels inside `Border` examples; they were unreadable in the dark theme variant.


src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,182 @@
...
+                if (_registered)
+                    return;
+                _registered = true;
+            }
+
+            try
+            {
+                Register();
+            }
+#pragma warning disable CA1031 // registration touches Avalonia internals reflectively; never fail app startup
```
jinek: I don't understand why we do not allow exceptions here

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,182 @@
...
+                Assembly avalonia = typeof(InterpolatingAnimator<>).Assembly;
+                Type gradientAnimator = avalonia.GetType("Avalonia.Animation.Animators.GradientBrushAnimator", true)!;
+                object instance = Activator.CreateInstance(gradientAnimator, true)!;
+                MethodInfo interpolate = gradientAnimator.GetMethod("Interpolate",
+                    new[] { typeof(double), typeof(IGradientBrush), typeof(IGradientBrush) })!;
+
+                return (Func<double, IGradientBrush, IGradientBrush, IGradientBrush>)Delegate.CreateDelegate(
+                    typeof(Func<double, IGradientBrush, IGradientBrush, IGradientBrush>), instance, interpolate);
+            }
+#pragma warning disable CA1031 // binding to an Avalonia internal; degrade to discrete switching on failure
```
jinek: Same here. If we expect some signature to change or methods been removed then we should await specific exception or check MethodInfo for null etc

mstrobel: This was my rationale:

We have tests verifying that not only was the animator registered, but it is functioning properly. If it ever breaks, the tests will tell you.

However, this is a low-impact feature in a high-impact class (registration is initiated in `LineBrush.cctor`). If the exception propagates, it forces you to deal with the this before addressing more critical integration issues with a future Avalonia release.

But I can certainly remove the exception handler if you prefer.

src/Consolonia.Core/Drawing/BrushExtensions.cs
```diff
@@ -1,135 +1,211 @@
...
-            if (after == null) return before.Color;
+            // Pad to the first/last stop color outside the covered offset range.
+            if (before == null)
+                return ApplyOpacity(after.Color, brush.Opacity);
+            if (after == null)
+                return ApplyOpacity(before.Color, brush.Opacity);
+            if (before == after)
+                return ApplyOpacity(before.Color, brush.Opacity);
+
+            double ratio = (position - before.Offset) / (after.Offset - before.Offset);
```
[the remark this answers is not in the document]
mstrobel: Fixed.

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,182 @@
...
+        ///     multiple times; only the first call has an effect, and any failure is logged rather than thrown.
+        /// </summary>
+        public static void EnsureRegistered()
+        {
+            lock (RegistrationGate)
+            {
+                if (_registered)
+                    return;
+                _registered = true;
+            }
```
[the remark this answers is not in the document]
mstrobel: Fixed

mstrobel:
> 🧹 Nitpick comments (1)
> > src/Tests/Consolonia.Core.Tests/BrushExtensionsTests.cs (1)> `31-205`: _⚡ Quick win_
> > > **Add coverage for unsorted stops and brush opacity.**
> > > These are stated behavior contracts, but this suite does not currently assert either one explicitly. Please add focused tests for both.

Tests added and passing.

mstrobel:
> Write docstrings for the functions missing them to satisfy the coverage threshold.

Added API documentation to all new non-test methods that were lacking them.

mstrobel:
> `49-68`: _💤 Low value_
> 
> **`Thread.MemoryBarrier()` is redundant inside the lock.**
> 
> The `lock` statement already issues full acquire/release barriers. The extra barrier has no effect here.
> 
> Suggested removal

True. I had a brain fart and was treating `_registered` like a field in the Avalonia class being accessed through reflection. In that case, it could have made a material difference outside of the `LineBrushAnimator` class if the write got reordered to occur before all the list changes were committed.

Not sure it's worth going through the inspection process again just to fix it, though.

furesoft (dismissed):

src/Consolonia.Core/Drawing/BrushExtensions.cs
```diff
@@ -1,135 +1,233 @@
...
-        private static Color InterpolateColor(IGradientBrush brush, double relativePosition)
+        /// <summary>
+        ///     Apply the gradient's spread method and evaluate its stops at the given parameter.
+        /// </summary>
+        private static Color GetGradientColor(IGradientBrush brush, double position)
         {
+            IReadOnlyList<IGradientStop> stops = brush.GradientStops;
+
+
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalse
```
jinek, line 140: Why can it be null?

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
jinek: I understand the intention to make a failure independent but we don't know the type of exception or it's impact and how critical it is. For instance, it could be OOM or any security relevant exception expected by the user.
Also, importance of the feature we don't know in general.

I'm proposing either to change the initialization to be independently "fail-able" or just remove the catch - we don't prioritize failures anywhere else, as I see similar to Avalonia.

BTW. I have noticed that there is a method `BaseBrushAnimator.RegisterBrushAnimator` - does not it simplify life anyhow?

jinek:
One of the check fails - it found duplicate in test files. you can ignore that file with `// DUPFINDER_ignore` comment at the top of a file @mstrobel

mstrobel:
In response to AI reviewer, `LineBrushAnimator` now additionally interpolates the underlying brush's `Opacity` and `Transform`.

Possible exceptions during registration are still logged, but now propagate as requested. Additional calls to the method have no effect.

furesoft (dismissed):

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,283 @@
...
+    {
+        // Reflected handle to Avalonia's internal GradientBrushAnimator.Interpolate, used to animate a wrapped
+        // gradient exactly the way Avalonia animates an unwrapped one.
+        private static readonly Func<double, IGradientBrush, IGradientBrush, IGradientBrush> GradientInterpolator =
+            CreateGradientInterpolator();
+
+        private static readonly object RegistrationGate = new();
+        private static bool _registered;
+
+#pragma warning disable CA1031
```
jinek: Now can restore CA1031 back

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,283 @@
...
+
+                return (Func<double, IGradientBrush, IGradientBrush, IGradientBrush>)Delegate.CreateDelegate(
+                    typeof(Func<double, IGradientBrush, IGradientBrush, IGradientBrush>), instance, interpolate);
+            }
+            catch (Exception e)
+            {
+                Logger.TryGet(LogEventLevel.Warning, LogArea.Animations)?.Log(null,
+                    "{Animator} could not bind the gradient interpolator: {Message}", nameof(LineBrushAnimator),
+                    e.Message);
+                return null;
```
jinek: here same? just re-throw?

jinek:
@mstrobel I have little time for this. I can take this PR over if you think makes sense.

mstrobel:
Works for me!

src/Consolonia.Controls/Brushes/LineBrushAnimator.cs
```diff
@@ -0,0 +1,283 @@
...
+                newTransform is TransformOperations newTransformOperations)
+            {
+                return new ImmutableTransform(
+                    TransformOperations.Interpolate(oldTransformOperations, newTransformOperations, progress).Value);
+            }
+
+            if (oldTransform is not null)
+                return new ImmutableTransform(oldTransform.Value);
+            
+            return null;
```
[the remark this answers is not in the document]
jinek: Explain this more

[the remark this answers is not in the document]
jinek: Just create a separate issue about this please

mstrobel:
FYI, I'm pretty sure I replicated the transform interpolation verbatim from one of Avalonia's existing animators (probably the gradient brush animator). So the behavior would be consistent with what they ship.

jinek:
Colleagues, please review one more time @mstrobel  @furesoft 
I guess there are no more objectives. One of the issues reported by coderabbit I've extracted to a separate issue for further investigation.

furesoft (approved):

PR #680, opened 2026-08-14
Fix Consolonia app lifetime isolation per test

### LINKED ISSUE #679: Tests lifetime
ConsoloniaAppTestBase can initialize the application only once per entire app domain. We should make it to teardown and spin up for each test.

https://github.com/Consolonia/Edit.NET/actions/runs/31764312730/job/94656923807?pr=61

floze-the-genius:
Closes #679.

## What changed

- Run `ConsoloniaAppTestBase` setup and teardown for every NUnit test instead of once per fixture/process.
- Keep all application lifetimes on one serialized background UI thread, reset Avalonia's test dispatcher state before and after each lifetime, and clean up the lifetime, console, and locator scope on that same thread.
- Run the gallery navigation setup per test and make the Expander/Fonts fixtures independent of execution order.
- Add a regression fixture proving that consecutive tests receive distinct `Application.Current` instances.

## Why the dedicated thread/reset sequence

Avalonia dispatchers and cached render resources retain thread affinity. Recreating the app on arbitrary thread-pool workers either reuses a shut-down dispatcher or eventually binds rendering to a different thread. The serialized app thread plus Avalonia's own internal unit-test reset hooks keeps each application isolated without moving its UI loop between threads.

## Validation

- Regression before the production change: 1 passed, 1 failed because both tests observed the same application instance.
- Regression after the change: 2/2 passed.
- `dotnet build --no-restore --configuration release`: 0 warnings, 0 errors.
- `LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 DOTNET_ROLL_FORWARD=Major dotnet test --no-restore`: Core 306 passed / 1 skipped; Gallery 47 passed / 2 skipped.

The locale is explicit because the existing calendar assertions expect US date formatting and week layout. `DOTNET_ROLL_FORWARD=Major` is only needed on the local validation host, which has the .NET 10 runtime; CI installs the repository's .NET 8 target directly.


jinek:
@floze-the-genius Thank you for the contribution. I'm reviewing it.

If you pick up any other tasks, please assign them first. Or otherwise we do double work https://github.com/Consolonia/Consolonia/pull/681

src/Consolonia.NUnit/Assembly.cs
```diff
@@ -1,3 +1,3 @@
-using System;
-
-[assembly: CLSCompliant(false)] //todo: should we make it compliant?
\ No newline at end of file
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Consolonia.Gallery.Tests")]
```
jinek, line 3: Why do we need internals there?

src/Consolonia.NUnit/ConsoloniaAppTestBase.cs
```diff
@@ -87,26 +156,75 @@ await Task.Run(async () =>
...
+        private static async Task AwaitWhileWorkerIsRunning(Task operation, Task workerTask)
+        {
+            Task completedTask = await Task.WhenAny(operation, workerTask).ConfigureAwait(true);
+            if (completedTask == workerTask)
+            {
+                await workerTask.ConfigureAwait(true);
+                throw new InvalidOperationException("Application lifetime ended during test setup.");
+            }
+
+            await operation.ConfigureAwait(true);
```
jinek, line 220: I don't understand logic of this method. At line 213 "operation" task was already awaited (we know this because otherwise it would throw at line 217). Why to await it again at line 220?

src/Consolonia.NUnit/ConsoloniaAppTestBase.cs
```diff
@@ -87,26 +156,75 @@ await Task.Run(async () =>
...
-            _scope.Dispose();
-            _scope = null;
-            UITest.Dispose();
+            Cleanup(() => UITest?.Dispose());
             UITest = null;
-            await _disposeTaskCompletionSource.Task.ConfigureAwait(true);
+            Cleanup(() => ResetDispatcher("ResetForUnitTests"));
+            Cleanup(() => _scope?.Dispose());
+            _scope = null;
+            Cleanup(() => ResetDispatcher("ResetBeforeUnitTests"));
```
jinek, line 205: just a guess: may be chain of try{} finally{ try{}finally{}} would look simplier than Cleanup declaration + exception capturing.

src/Consolonia.NUnit/ConsoloniaAppTestBase.cs
```diff
@@ -44,38 +98,53 @@ protected virtual AppBuilder CreateAppBuilder()
...
+                    _ = Dispatcher.CurrentDispatcher;
+                    _scope = AvaloniaLocator.EnterScope();
+                    _lifetime = ApplicationStartup.CreateLifetime(CreateAppBuilder(), Args);
+                    uiTest.SetupLifetime(_lifetime);
+                    setupTaskSource.TrySetResult();
+                    _lifetime.Start(Args);
+                }
+                finally
+                {
+                    CleanupTestApplication();
```
jinek, line 125: Why don't we do Cleanup from [TearDown] ?

jinek: It looks to me like a race between CleanupTestApplication and [TearDown] method. This cleanup, for instance, disposes the lifetime. Sametime [TearDown] method already passed the condition and tries to shutdown the same lifetime. @coderabbitai could you please validate my doubts?

jinek:
I've resurrected similar PR as we stuck with this one for some time. https://github.com/Consolonia/Consolonia/pull/681

